PR #1, opened 2017-12-05
Spring Framework Integration

andrewpowell:
This integration for the Spring Framework enables the usage of aspects to trace requests down a call stack.  Classes can either implement an interface or be annotated to identify themselves as available to the aspect for tracing.


aws-xray-recorder-sdk-spring/src/main/java/com/amazonaws/xray/spring/aop/AbstractXRayInterceptor.java
```diff
@@ -0,0 +1,102 @@
+package com.amazonaws.xray.spring.aop;
+
+import com.amazonaws.xray.AWSXRay;
+import com.amazonaws.xray.entities.Subsegment;
+import com.amazonaws.xray.exceptions.SegmentNotFoundException;
+import org.aspectj.lang.ProceedingJoinPoint;
+import org.aspectj.lang.annotation.Around;
+import org.aspectj.lang.annotation.Pointcut;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
```
jamesdbowman: As an SDK convention we have been relying on `org.apache.commons.logging.Log` for our log framework, only because it is bundled with the AWS SDK already, and keeps the logging consistent between the two. I'd be open to switching to `slf4j` across the SDK if a good case could be made for diverging from the AWS SDK, though.

aws-xray-recorder-sdk-spring/src/main/java/com/amazonaws/xray/spring/aop/AbstractXRayInterceptor.java
```diff
@@ -0,0 +1,102 @@
...
+
+    private final Logger LOGGER = LoggerFactory.getLogger(this.getClass());
+
+    protected Object processXRayTrace(ProceedingJoinPoint pjp) throws Throwable {
+        boolean endSegment = false;
+        try {
+            if (AWSXRay.getCurrentSegment() != null) {
+                LOGGER.trace("Current segment exists");
+            }
+        } catch (SegmentNotFoundException snfe) {
```
jamesdbowman: The segment not found cases should be passed to the current recorder's `ContextMissingStrategy`. We try to protect users from inadvertent tracing context loss (happens most often when trying to trace multi-threaded operations). In order to protect users from losing their tracing context, we must by default throw an exception when a user tries to create a subsegment with no segment in progress. Of course, users may override this if they desire. But the default behavior should pass the exception onto the context missing strategy and allow that strategy to create a default segment, if it so desires.

aws-xray-recorder-sdk-spring/src/main/java/com/amazonaws/xray/spring/aop/AbstractXRayInterceptor.java
```diff
@@ -0,0 +1,102 @@
...
+            AWSXRay.beginSegment(pjp.getClass().getSimpleName());
+            endSegment = true;
+        }
+
+        try {
+            Subsegment subsegment = AWSXRay.beginSubsegment(pjp.getSignature().getName());
+            subsegment.setMetadata(this.generateMetadata(pjp, subsegment));
+            return XRayInterceptorUtils.conditionalProceed(pjp);
+        } catch (Exception e) {
+            LOGGER.error(e.getMessage());
```
jamesdbowman: I think we should avoid re-logging any exceptions that may pass through customer code, as they likely will be logging it already.

aws-xray-recorder-sdk-spring/src/main/java/com/amazonaws/xray/spring/aop/XRayJDBCInterceptor.java
```diff
@@ -0,0 +1,51 @@
...
+    private void queryExecution(){}
+
+
+    public Map<String, Map<String, Object>> generateMetadata(ProceedingJoinPoint pjp, Subsegment subsegment) throws Exception{
+        final Map<String, Map<String, Object>> metadata = new HashMap<>();
+        final Map<String, Object> classInfo = new HashMap<>();
+        classInfo.put("Class", pjp.getTarget().getClass().getSimpleName());
+        metadata.put("ClassInfo", classInfo);
+        String arg = (String) pjp.getArgs()[0];
+        subsegment.putSql("SQL Statement",arg);
```
jamesdbowman: We've reserved the keyword `sanitized_query` in our segment schema for such cases, would it be possible to sanitize the query of any literal values and use that keyword here? Perhaps we could also provide a `query` only option (non-sanitized) but we should draw user's attention to the fact that they are adding sensitive information to their traces.

andrewpowell:
Thx for the feedback.  I've updated the code with the suggestions and committed for review.

aws-xray-recorder-sdk-spring/pom.xml
```diff
@@ -0,0 +1,84 @@
...
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-compiler-plugin</artifactId>
+                <configuration>
+                    <compilerVersion>1.8</compilerVersion>
+                    <fork>true</fork>
+                    <source>1.8</source>
+                    <target>1.8</target>
+                </configuration>
+            </plugin>
```
jamesdbowman: could you add the Javadoc plugin to align this with other submodule poms?

```xml
        <groupId>org.apache.maven.plugins</groupId>
        <artifactId>maven-javadoc-plugin</artifactId>
        <version>2.10.4</version>
```

Additionally, it may be useful to consider adding another `<group>` tag in the main `pom.xml` to ensure these `com.amazonaws.xray.spring.*` classes are documented together in a group.

aws-xray-recorder-sdk-spring/pom.xml
```diff
@@ -0,0 +1,84 @@
...
+            <version>1.8.11</version>
+            <scope>compile</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.springframework.data</groupId>
+            <artifactId>spring-data-commons</artifactId>
+            <version>2.0.0.RELEASE</version>
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
```
jamesdbowman: Sorry if I'm missing something but it doesn't look like this dependency on `javax.servlet.servlet-api` is used in the submodule.

aws-xray-recorder-sdk-spring/pom.xml
```diff
@@ -0,0 +1,84 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd" xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>com.amazonaws</groupId>
+        <artifactId>aws-xray-recorder-sdk-pom</artifactId>
+        <version>1.2.1</version>
```
jamesdbowman: Sorry for the trouble but the version has bumped to `1.2.2` since this was posted, would it be possible to increment?

.gitignore
```diff
@@ -5,3 +5,6 @@
 .project
 .classpath
 .DS_Store
+
+*.iml
+/.idea
```
jamesdbowman: nitpick but it may mess up other users' editors - can we add newlines at EOF where missing?

aws-xray-recorder-sdk-spring/pom.xml
```diff
@@ -0,0 +1,84 @@
...
+        <dependency>
+            <groupId>org.springframework</groupId>
+            <artifactId>spring-aspects</artifactId>
+            <version>${spring.version}</version>
+            <scope>compile</scope>
+        </dependency>
+        <dependency>
+            <groupId>com.amazonaws</groupId>
+            <artifactId>aws-xray-recorder-sdk-core</artifactId>
+            <version>1.2.0</version>
```
jamesdbowman: Using `<version>${awsxrayrecordersdk.version}</version>` will allow this submodule to directly depend on the latest version of the core module.

aws-xray-recorder-sdk-spring/src/main/java/com/amazonaws/xray/spring/aop/AbstractXRayInterceptor.java
```diff
@@ -0,0 +1,95 @@
...
+        return this.processXRayTrace(pjp);
+    }
+
+    protected Object processXRayTrace(ProceedingJoinPoint pjp) throws Throwable {
+        boolean endSegment = false;
+        try {
+            if (AWSXRay.getCurrentSegment() != null) {
+                logger.trace("Current segment exists");
+            }
+        } catch (SegmentNotFoundException snfe) {
```
jamesdbowman: Sorry if my previous comment regarding context missing strategies was unclear. The logic inside the `getCurrentSegment()` function should reference the context missing strategy as needed - and will only throw the `SegmentNotFoundException` if the strategy calls for it.

```java
    public Segment getCurrentSegment() {
        Optional<Segment> segment = getCurrentSegmentOptional();
        if (segment.isPresent()) {
            return segment.get();
        }
        contextMissingStrategy.contextMissing("No segment in progress.", SegmentNotFoundException.class);
        return null;
    }
```

So, there should be no need to catch exceptions here. And the `boolean endSegment` shouldn't be required either. We can call `getSegment()` and compare the value against `null` to determine whether or not a subsegment should be created here.

andrewpowell:
Incorporated latest round of feedback

aws-xray-recorder-sdk-apache-http/pom.xml
```diff
@@ -1,5 +1,5 @@
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-  xsi:schemaLocation="http://mapven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+<project xmlns="http://maven.apache.org/POM/4.0.0"
```
jamesdbowman: +1 thanks for fixing these across the board!! 😁

aws-xray-recorder-sdk-spring/src/main/java/com/amazonaws/xray/spring/aop/AbstractXRayInterceptor.java
```diff
@@ -0,0 +1,98 @@
...
+     */
+    @Around("xrayTracedClasses() || xrayEnabledClasses()")
+    public Object traceAroundMethods(ProceedingJoinPoint pjp) throws Throwable {
+        return this.processXRayTrace(pjp);
+    }
+
+    protected Object processXRayTrace(ProceedingJoinPoint pjp) throws Throwable {
+        Segment segment = getCurrentSegment();
+        if (segment == null) {
+            AWSXRay.beginSegment(pjp.getSignature().getName());
```
jamesdbowman: I noticed that this `beginSegment` remains while the `endSegment` logic was taken out. I'm still not sure we can get away with beginning the segments if one doesn't exist - I think we'd have to leave that up to the `SegmentContext` / `ContextMissingStrategy` chosen by the programmer.

andrewpowell: fixed and pushed

PR #3, opened 2017-12-28
Bump to Version 1.3.0

andrewpowell:
Bumping the version up to 1.3.0


haotianw465:
Hi, version bump should be associated to a single commit. Looks like this PR has some commit entries that are irrelevant to the version bump. Could you squash them to a single clean commit?

PR #10, opened 2018-01-21
Documentation fix for defaultRecorder method

JaredHatfield:
The documentation was calling the AWSXRayRecorderBuilder.defaultBuilder(); method but that one does not exist.  The method name is actually defaultRecorder


luluzhao:
@JaredHatfield , thank you for your fix. We merged your PR.

Issue #26, opened 2018-09-04
Workarounds and visibility on IOException: Message too long


steven-aerts:
We are running into the well known (#4) problem that sometimes a segment is too big to fit in a UDP message (65,507 bytes).  For example because it contains a big sql statement or an exception.

The problem is however that if these issues happen sporadically they are very difficult to debug as there is almost no visibility.

We see a few possible improvements which can be made easily:

* _Enhance visibility_: For example by logging the segment when it is too big.
It could be as easy as doing something like this:
```diff
     }
 
-    private boolean sendData(byte[] data) {
+    private boolean sendData(byte[] data, Entity entity) {
+        if (data.length > 65_527) {
+            logger.warn("(sub)segment is too big to send: " + entity.prettySerialize());
+            return false;
+        }
         DatagramPacket packet = new DatagramPacket(sendBuffer, DAEMON_BUF_RECEIVE_SIZE, config.address);
```
(Btw: it also puzzles us why `DAEMON_BUF_RECEIVE_SIZE` is a lot bigger then 65_527)

* _Add retry behavior_ which tries to send the completed subsegments (so if one is too big, it does not break the others).  We for example currently quickly implemented a `RetryingEmitter` which works like this:

```java
public class RetryingEmitter extends UDPEmitter {
    private static final Log logger = LogFactory.getLog(MethodHandles.lookup().lookupClass());
    private final StreamingStrategy streamingStrategy;
    public RetryingEmitter(StreamingStrategy streamingStrategy) throws SocketException {
        super();
        this.streamingStrategy = streamingStrategy;
    }
    public boolean sendSegment(Segment segment) {return trySplitOnFailure(segment, super::sendSegment);}
    public boolean sendSubsegment(Subsegment subsegment) {return trySplitOnFailure(subsegment, super::sendSubsegment);}
    <T extends Entity> boolean trySplitOnFailure(T entity, Predicate<T> send) {
        boolean ok = send.test(entity);
        if (!ok) {
            logger.info("send of entity " + entity.getId() + " failed, retrying.");
            streamingStrategy.streamSome(entity, this);
            ok = send.test(entity);
            if (!ok) {
                logger.warn("Unable to send entity " +  entity.getId() + " after split. size: " +
                        entity.serialize().length() + ", content: " + entity.prettySerialize());
            }
        }
        return ok;
    }
}
```
This is of course rather naive, as a lot of errors are still send.  But it helps us to debug these kinds of issues.

[...]

PR #32, opened 2018-09-28
GH-26 log entity when UDP emit failed

steven-aerts:
Implements first enhancement proposed in #26.


aws-xray-recorder-sdk-core/src/main/java/com/amazonaws/xray/emitters/UDPEmitter.java
```diff
@@ -58,17 +58,17 @@ public boolean sendSubsegment(Subsegment subsegment) {
...
 
-    private boolean sendData(byte[] data) {
+    private boolean sendData(byte[] data, Entity entity) {
         DatagramPacket packet = new DatagramPacket(sendBuffer, DAEMON_BUF_RECEIVE_SIZE, config.address);
         packet.setData(data);
         try {
             logger.debug("Sending UDP packet.");
             daemonSocket.send(packet);
         } catch (IOException e) {
-            logger.error("Exception while sending segment over UDP.", e);
```
luluzhao: Could you do not delete this one? From my understanding, there should have two log messages. One is log error showing the exception what happened through sending data. The second one should be log debug showing which segments/subsegments fail to be sent out.

aws-xray-recorder-sdk-core/src/main/java/com/amazonaws/xray/emitters/UDPEmitter.java
```diff
@@ -58,17 +58,17 @@ public boolean sendSubsegment(Subsegment subsegment) {
...
+            logger.error("Exception while sending entity over UDP: " + entity.prettySerialize(), e);
```
luluzhao: As I said in the first comment, could you change this one to only print the segments/subsegments?

aws-xray-recorder-sdk-core/src/main/java/com/amazonaws/xray/emitters/UDPEmitter.java
steven-aerts: I kept the original message and moved the printing of the entity to the info log level.
Is that more acceptible to you?

Moving it to debug level will also hit the debug statement in `com.amazonaws.xray.emitters.UDPEmitter#sendSegment` which will print all possible traces, so then you also have no clue anymore which one was too big and failed.

aws-xray-recorder-sdk-core/src/main/java/com/amazonaws/xray/emitters/UDPEmitter.java
steven-aerts: Only printing the current segment/sub-segment id won't help us much I think, as the problem is in our use case typically in an underlying sub-segment.

aws-xray-recorder-sdk-core/src/main/java/com/amazonaws/xray/emitters/UDPEmitter.java
```diff
@@ -58,17 +58,20 @@ public boolean sendSubsegment(Subsegment subsegment) {
...
+    private boolean sendData(byte[] data, Entity entity) {
         DatagramPacket packet = new DatagramPacket(sendBuffer, DAEMON_BUF_RECEIVE_SIZE, config.address);
         packet.setData(data);
         try {
             logger.debug("Sending UDP packet.");
             daemonSocket.send(packet);
         } catch (IOException e) {
             logger.error("Exception while sending segment over UDP.", e);
+            if (logger.isInfoEnabled()) {
+                logger.info("Failed to send the following entity: " + entity.prettySerialize());
```
haotianw465: Segment/subsegment documentation has `annotation` and `metadata` key which could contain sensitive customer data. Right now the only documentation printing on log file is debug mode so normally no content would be written down in production. This might not suit most use cases of the customers. Any reason not using `DEBUG` here?

steven-aerts: Is there any other way how we can give a use a pointer which entity was too big?
Because this is something you want to know beyond the debug level I think.
If you enable DEBUG, you see all xray traces passing by, as this trace want to give the end use a clue why it failed.

steven-aerts: @haotianw465 I understand you security concerns.  I will update the patch to print the segment id in the error trace.

Issue #35, opened 2018-10-12
ClassCastException when invoking beginSubsegment


denyskonakhevych:
I experience following exception when try to invoke lambda directly with X-Ray tracing feature
```
java.lang.ClassCastException: com.amazonaws.xray.entities.SegmentImpl cannot be cast to com.amazonaws.xray.entities.Subsegment
at com.amazonaws.xray.contexts.LambdaSegmentContext.beginSubsegment(LambdaSegmentContext.java:53) ~[task/:?]
at com.amazonaws.xray.AWSXRayRecorder.beginSubsegment(AWSXRayRecorder.java:410) ~[task/:?]
at com.amazonaws.xray.handlers.TracingHandler.beforeRequest(TracingHandler.java:152) ~[task/:?]
at com.amazonaws.http.AmazonHttpClient$RequestExecutor.runBeforeRequestHandlers(AmazonHttpClient.java:770) ~[task/:?]
at com.amazonaws.http.AmazonHttpClient$RequestExecutor.doExecute(AmazonHttpClient.java:724) ~[task/:?]
at com.amazonaws.http.AmazonHttpClient$RequestExecutor.executeWithTimer(AmazonHttpClient.java:717) ~[task/:?]
at com.amazonaws.http.AmazonHttpClient$RequestExecutor.execute(AmazonHttpClient.java:699) ~[task/:?]
at com.amazonaws.http.AmazonHttpClient$RequestExecutor.access$500(AmazonHttpClient.java:667) ~[task/:?]
at com.amazonaws.http.AmazonHttpClient$RequestExecutionBuilderImpl.execute(AmazonHttpClient.java:649) ~[task/:?]
at com.amazonaws.http.AmazonHttpClient.execute(AmazonHttpClient.java:513) ~[task/:?]
at com.amazonaws.services.lambda.AWSLambdaClient.doInvoke(AWSLambdaClient.java:2654) ~[task/:?]
at com.amazonaws.services.lambda.AWSLambdaClient.invoke(AWSLambdaClient.java:2630) ~[task/:?]
at com.amazonaws.services.lambda.AWSLambdaClient.executeInvoke(AWSLambdaClient.java:1497) ~[task/:?]
at com.amazonaws.services.lambda.AWSLambdaClient.invoke(AWSLambdaClient.java:1473) ~[task/:?]
*** omitted handler code ***
[...]

PR #36, opened 2018-10-12
AXSJ-35: ClassCastException when invoking beginSubsegment

denyskonakhevych:
ClassCastException when invoking beginSubsegment

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/aws-xray-recorder-sdk-core/src/main/java/com/amazonaws/xray/contexts/LambdaSegmentContext.java
+++ b/aws-xray-recorder-sdk-core/src/main/java/com/amazonaws/xray/contexts/LambdaSegmentContext.java
@@ -37,7 +37,8 @@ public class LambdaSegmentContext implements SegmentContext {
         if (logger.isDebugEnabled()) {
             logger.debug("Beginning subsegment named: " + name);
         }
-        if (null == getTraceEntity()) { // First subsgment of a subsegment branch.
+        Entity current = getTraceEntity();
+        if (null == current) { // First subsegment of a subsegment branch.
             Segment parentSegment = null;
             if (LambdaSegmentContext.isInitializing(LambdaSegmentContext.getTraceHeaderFromEnvironment())) {
                 logger.warn(LAMBDA_TRACE_HEADER_KEY + " is missing a trace ID, parent ID, or sampling decision. Subsegment " + name + " discarded.");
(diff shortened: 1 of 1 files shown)


haotianw465:
Thank you for your contribution. This issue https://github.com/aws/aws-xray-sdk-java/issues/35 indeed needs a fix. Before merging this PR, I would like to make sure this change is the right behavior for the general use case.

When the SDK implements lambda function support it assumes inside a lambda function only subsegments are stored in tracing context since lambda container is the one generating the parent segment which the SDK doesn't have access to. This has been working fine so far. Could you provide more information about your use case so we can confirm the right solution to deal with another segment stored in lambda context in general?

denyskonakhevych:
Yes, you are right. I started new segment within request handling code in lambda.
When I removed creating of segment before calling lambda - it created new facade segment and added subsegment.
Nevertheless, I thought it would be a good point to separate logic regarding audit log handling in some another segment (like something separated into common logic for multiple services and not considered as a part of particular requests). If it's out of concept of lambda treating xray - it could be moved to separate rest service and add exception, like subsegment cannot be found.
If so - we have some inconsistent logic with ThreadLocalSegmentContext, where it's possible to add subsegment to not null segments started by user.
Could you please advice on this or if you need more info from me - feel free to ask

denyskonakhevych:
Now I see another problem. 
So I have lambda handler, that implements com.amazonaws.services.lambda.runtime.RequestHandler.
Inside my code I don't create any segment or subsegments. Than TracingHandler comes into play.
At 164 line it execute recorder.beginSubsegment(serviceName). At this point trace entity is null, so it creates facade segment and use it as a parent segment for new subsegment.
But at 175 line of TracingHandler it tries to retrieve current segment, which is null. As a result - trace header is not passed to executing service.
So it looks like I don't have segment when handling lambda. Did I miss something and it should be there?

denyskonakhevych:
And even if I have segment generated by AWS and I can get it inside lambda. Should it be also subsegment generated by AWS?

luluzhao:
Hello @denyskonakhevych, for lambda integration, xray sdk is only responsible for creating subsegment inside lambda function by current design. Segment is generated and sent out to xray service by lambda container and our xray sdk doesn't have access to it. In lambda side, they write trace header information into environment variable, then xray sdk uses this trace header information to create a facade segment and all subsegments generated by sdk will be added under the facade segment. However, the facade segment won't be sent to xray service and only subsegments do.

Here I attached a public link which contains the concepts of segment and subsegment: https://docs.aws.amazon.com/xray/latest/devguide/xray-concepts.html. Feel free to ask me if you have any questions about these definitions. 

What's more, could you tell us more about your use case for creating a segment inside lambda function? If it is valid, we have to update our sdk according to provide this feature.

denyskonakhevych:
Hi @luluzhao. 
In my case when handling lambda request I try to execute http request with TracingHandler. Trace entity is null so it creates facade segment and subsegment for request. However, after that tracing handler tries to retrieve current segment (recorder.getCurrentSegment()) which is null and do not pass header to request.
If I try to create segment inside lambda function - I get class cast exception, since TracingHandler creates subsegment based on subsegment only.

luluzhao:
Hello @denyskonakhevych , `TracingHandler` is specifically for tracking aws downstream calls. [Here is the Reference](https://docs.aws.amazon.com/xray/latest/devguide/xray-sdk-java-awssdkclients.html). For general outgoing http request, I would suggest you using `HttpClientBuilder` xray sdk provided. Here I attached a link for [example](https://docs.aws.amazon.com/xray/latest/devguide/xray-sdk-java-httpclients.html).

Lordevss:
Run

denyskonakhevych:
I looked at this problem once again and it works with sub-segments.
Sorry for disturbing you and thank you for clarification

PR #47, opened 2018-12-10
Improves reservoir to avoid a race condition and to be more fair

adriancole:
This re-uses the implementation we just made in Brave based on feedback
on yours. Enjoy!

See https://github.com/openzipkin/brave/pull/819

PR #52, opened 2019-01-06
Initializes Reservoir reset setting  eagerly.

chang-chao:
As in the [javadoc](https://javadoc.scijava.org/Java9/java/lang/System.html#nanoTime--)  states, `System.nanoTime()` may return a negative value, the current implementation  of Reservoir may initially set the reset time to 0, when this value is compared to `System.nanoTime()` to determine max traces per decisecond,  an IndexOutOfBoundsException may occur in case of a negative nano time value.

This bug is introduced by the [this commit](https://github.com/aws/aws-xray-sdk-java/pull/47/commits/1c327137a47e31a28c3155a5852fe8d6052a5d48).


chang-chao:
@chrisradek Thanks.
As all the checks have  passed,the merging is blocked because approving review is required by reviewers with write access, so would you please approve it?

chrisradek:
@chang-chao 
Thank you for your contribution! Merging your changes now.

Issue #43, opened 2018-11-26
Add and document X-Ray tracing for AWS SDK v2


ashleymercer:
## Expected Behavior

The AWS SDK team has now released [version 2](https://github.com/aws/aws-sdk-java-v2) of the Java SDK. Previously it was possible to write:

```java
AWSLambda lambda = AWSLambdaClientBuilder
      .standard()
      .withRequestHandlers(new TracingHandler(recorder))
      .build()
```

but the `RequestHandler` API has been replaced in v2 with a new `ExecutionInterceptor` API for wrapping client requests:

https://github.com/aws/aws-sdk-java-v2/blob/master/core/sdk-core/src/main/java/software/amazon/awssdk/core/interceptor/ExecutionInterceptor.java

This means that, as far as I can see, it's currently not possible to add X-Ray traces to AWS calls using the v2 SDK.

## Current Behavior

I can't see an equivalent implementation in the v2 SDK anywhere - assuming it would use the new `ExecutionInterceptor` API - or if there is one, I can't see it documented in [the developer guide](https://docs.aws.amazon.com/sdk-for-java/v2/developer-guide/welcome.html).

## Possible Solution
Add and document an implementation of `ExecutionInterceptor` that traces AWS client requests.

## Steps to Reproduce (for bugs)
N/A

## Context
Required for parity with the 1.11 SDK.

## Your Environment
* AWS Java SDK version used: 2.1.0
* JDK version used: 8

awssandra:
Hi ashleymercer,

We are aware of the current issues and are actively working with the AWS SDK team to support X-Ray. We plan to have official support soon. Thank you for your patience, and stay tuned.

Thanks,
Sandra

haotianw465:
AWS SDK for Java V2 is supported in 2.2.x. You can see the usage guide at the bottom of https://docs.aws.amazon.com/xray/latest/devguide/xray-sdk-java.html. We are working on improving the docs. Feel free to put extra feedback for the doc.

PR #56, opened 2019-01-23
Adds support for instrumenting the AWS SDK for Java V2

chrisradek:
*Issue #, if available:*
#43 

*Description of changes:*
Adds 2 new artifacts for instrumenting the AWS SDK for Java V2.

The `aws-xray-recorder-sdk-aws-sdk-v2` artifact exposes a `TracingInterceptor` class. An instance of this class can be passed into any AWS SDK V2 service client when it is being built:
```java
DynamoDbAsyncClient client =  DynamoDbAsyncClient.builder()
    .region(Region.US_WEST_2)
    .overrideConfiguration(ClientOverrideConfiguration.builder()
        .addExecutionInterceptor(new TracingInterceptor())
        .build()
    ).build();
```

The `aws-xray-recorder-sdk-aws-sdk-v2-instrumentor` artifact allows for easy instrumentation of all AWS SDK V2 service clients. When included as a dependency, the AWS SDK will automatically inject the `TracingInterceptor` into each service client.

Both artifacts will require version `2.2.0` of the `aws-sdk-java` artifact in order to run.

I haven't dealt with maven-managed dependencies much. Currently, neither artifact will work unless the AWS SDK V2 is also a dependency of the consuming package. This allows the customer to specify which version of the AWS SDK they wish to use. If testing shows Maven will prefer the version of an artifact that is closest to the root pom.xml (and gradle works similarly) then it might make sense to explicitly depend on the AWS SDK in the new artifacts as well.


aws-xray-recorder-sdk-aws-sdk-v2/pom.xml
```diff
@@ -0,0 +1,70 @@
...
+            <groupId>org.skyscreamer</groupId>
+            <artifactId>jsonassert</artifactId>
+            <version>1.3.0</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>software.amazon.awssdk</groupId>
+            <artifactId>aws-sdk-java</artifactId>
+            <version>2.2.0</version>
+            <scope>provided</scope>
```
haotianw465: I looked up `provided` on Maven official doc but not quite clear what does this mean if user application specifies aws-sdk-java-bom with newer version or older version.

dmytro-bekuzarov:
If I use Java SDK v2 only, I got the following error when I try to use  `AWSXRayServletFilter`
```
java.lang.NoClassDefFoundError: com/amazonaws/services/xray/model/GetSamplingRulesRequest
	at com.amazonaws.xray.strategy.sampling.pollers.RulePoller.pollRule(RulePoller.java:64)
	at com.amazonaws.xray.strategy.sampling.pollers.RulePoller.lambda$start$0(RulePoller.java:46)
```
According to dependency tree, the x-ray sdk relies on AWS SDK 1.x version 
Can it be used using Java SDK v2 only?
P.S. I see there is a module `xray` under `software.amazon.awssdk`, what's its current status?

PR #57, opened 2019-01-23
Clear trace entities from asynchronous thread.

steven-aerts:
Previously a segment stayed on the worker thread when a servlet was processed asynchronously.
Which means that the next user of this thread will encounter this segment.
This typically results in traces like:

  Beginning new segment while another segment exists in the segment context.


chrisradek:
@steven-aerts 
Thanks for submitting this pull request as well! I'll take a look at this and #58 and respond back here soon!

aws-xray-recorder-sdk-core/src/test/java/com/amazonaws/xray/javax/servlet/AWSXRayServletFilterTest.java
```diff
@@ -53,6 +55,19 @@ private AWSXRayRecorder getMockRecorder() {
...
+        Mockito.doAnswer(a -> {
+            capturedEntity.set(AWSXRay.getTraceEntity());
+            return null;
+        }).when(chain).doFilter(request, response);
+
+        Mockito.when(request.getAttribute("com.amazonaws.xray.entities.Entity"))
+                .thenAnswer(a -> Objects.requireNonNull(capturedEntity.get()));
+        return chain;
+    }
+
```
chrisradek: @steven-aerts 
Thanks for updating these tests! Can you add one more that checks the case where `getAsyncContext()` fails? I think something like the below would cover it.

Otherwise, looks good!

```suggestion
@Test
    public void testAsyncServletRequestWithCompletedAsync() throws IOException, ServletException {
        AWSXRayServletFilter servletFilter = new AWSXRayServletFilter("test");

        HttpServletRequest request = Mockito.mock(HttpServletRequest.class);
        Mockito.when(request.getRequestURL()).thenReturn(new StringBuffer("test_url"));
        Mockito.when(request.getMethod()).thenReturn("TEST_METHOD");
        Mockito.when(request.isAsyncStarted()).thenReturn(true);
        Mockito.when(request.getAsyncContext()).thenThrow(IllegalStateException.class);

        HttpServletResponse response = Mockito.mock(HttpServletResponse.class);

        FilterChain chain = mockChain(request, response);

        AsyncEvent event = Mockito.mock(AsyncEvent.class);
        Mockito.when(event.getSuppliedRequest()).thenReturn(request);
        Mockito.when(event.getSuppliedResponse()).thenReturn(response);

        servletFilter.doFilter(request, response, chain);
        Assert.assertNull(AWSXRay.getTraceEntity());

        Mockito.verify(AWSXRay.getGlobalRecorder().getEmitter(), Mockito.times(1)).sendSegment(Mockito.any());
    }
```

steven-aerts:
@chrisradek thanks for the extra test, I think it does nicely what you suggest.

Thanks,

  Steven

PR #58, opened 2019-01-25
GH-26 RetryingUDPEmitter

steven-aerts:
An emitter which will try to stream the entity when its send failed.

A possible implementation for step 2 of #26.


chrisradek:
@steven-aerts 
Thank you for submitting this pull request! I'll try to take a close look and respond to this over the next couple days. With the changes to the udp emitter we want to make sure it's completely backwards-compatible, and really appreciate the context you provided in #26

luluzhao:
@steven-aerts thank you so much for contributing and sorry for the late response. I read your PR and I have a question about why you create another class `RetryingUDPEmitter`? Could you tell me more about it?

steven-aerts:
Hi @luluzhao ,

My first preference was/is that the retry behaviour was the standard behaviour of the `UDPEmitter`.
But @haotianw465  [advised](https://github.com/aws/aws-xray-sdk-java/issues/26#issuecomment-420792901) against it.  So I created the subclass.  But I would prefer this is the defaullt behaviour.

luluzhao:
@steven-aerts Travis CI tests are failing. Could you look into it and then fix it? Thank you.

steven-aerts:
I agree that subsegment streaming seems a lot better.

Will reopen this PR (or a similar one) when we see the exception happen in this new code.

Thanks

Issue #77, opened 2019-06-26
Add testing for JDK 11


chanchiem:
In pursuit of adding X-Ray support for JDK 11, we need to add JDK 11 in the testing suite.

PR #78, opened 2019-06-26
Add testing for JDK11 (#77)

chanchiem:
*Issue #, if available:*
#77 

*Description of changes:*
Add OpenJDK11

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,9 @@ addons:
 
 jdk:
   - openjdk8
+  - openjdk11
   - oraclejdk8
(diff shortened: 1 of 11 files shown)


luluzhao:
Would you mind adding the support for jdk 9 and 10? When the tests pass, we can be confident to say that we also support these two versions and it is only 11 is not working.

chanchiem:
Should we add 9 and 10? They're deprecated according to [this](https://en.wikipedia.org/wiki/Java_version_history) table.

PR #94, opened 2019-08-23
The method 'sendSubegment' of AWSXRay should have been 'sendSubsegment'

mb002270:
*Issue #, if available:*
No issue created.  

*Description of changes:*
This is fixing a typo in a method name.  sendSubegment vs sendSubsegment (missing the 's').

I marked 'sendSubegment' as deprecated and added a new 'sendSubsegment' method.
I made the old 'sendSubegment' call the new 'sendSubsegment' for backwards compatibility.
Since this is an interface change that is breaking (if the typo was just fixed) I figured the best way to handle this would be to mark the old method as deprecated, but keep it around.


chanchiem:
Thank you very much for your contribution!

Issue #45, opened 2018-12-05
Remove hard coded spring data dependency from com.amazonaws.xray.spring.aop.AbstractXRayInterceptor class, Project is not usable for non spring-data-jpa application 


ssrinivasulu:
Below code in AbstractXRayInterceptor is introducing spring-data-jpa library dependency and causing the library not to be used without adding "spring-data-jpa" the library to the project.  

    @Pointcut("execution(public !void org.springframework.data.repository.Repository+.*(..))")
    protected void springRepositories() {
    }

luluzhao:
Hello @ssrinivasulu , sorry for the late response. Here I want to confirm that the issue you are currently facing caused by our `pom.xml` file doesn't have `spring-data-jpa` jar included, is that correct?

ssrinivasulu:
Yes, that's correct. I would remove the dependency from code instead of adding the spring-data-jpa jar file to the aws-xray-sdk-java.

tmuldoon:
I've also run into this issue. 

We don't use spring-data-jpa and as such, we cannot seem to use the AbstractXRayInterceptor class, given the Pointcut reference...
```
    @Pointcut("execution(public !void org.springframework.data.repository.Repository+.*(..))")
    protected void springRepositories() {
    }
```
... results in the following exception...
```
java.lang.IllegalArgumentException: warning no match for this type name: org.springframework.data.repository.Repository [Xlint:invalidAbsoluteTypeName]
```

eric-swann-q2:
👆 Same here

saki95:
I have the same issue . Could anybody provide me with the solution . thanks

kgrodzicki:
> I have the same issue . Could anybody provide me with the solution . thanks

We have just copied 'AbstractXRayInterceptor' into project, deleted the @Pointcut("execution(public !void org.springframework.data.repository.Repository+.*(..))") and used this class as a base for XRayInterceptor. And it works like a charm, hope this helps.

XRayInterceptor:
`
import org.aspectj.lang.annotation.Aspect;
import org.aspectj.lang.annotation.Pointcut;
import org.springframework.stereotype.Component;

@Aspect
@Component
public class XRayInterceptor extends AbstractLeanXRayInterceptor {

    @Override
    @Pointcut("@within(com.amazonaws.xray.spring.aop.XRayEnabled)")
    public void xrayEnabledClasses() {
        // this defines which classes are enriched with X-Ray
    }

}

`

[...]

PR #115, opened 2019-12-17
Added interceptor without spring data dependency

dekelpilli:
*Issue #, if available:*
https://github.com/aws/aws-xray-sdk-java/issues/45

*Description of changes:*
Added another abstract interceptor, extended by the existing one. This shouldn't break backwards-compatibility, but gives the option to users who don't have spring-data-jpa on their classpath.

Note that I haven't been able to test this yet, will be doing that soon.

Edit: I won't be able to test this, as the original library itself doesn't work for my use case. I'll leave this open in case it's still useful for this quite old issue.

### DIFF
--- a/aws-xray-recorder-sdk-spring/src/main/java/com/amazonaws/xray/spring/aop/BaseAbstractXRayInterceptor.java
+++ b/aws-xray-recorder-sdk-spring/src/main/java/com/amazonaws/xray/spring/aop/BaseAbstractXRayInterceptor.java
@@ -71,18 +71,6 @@ public abstract class BaseAbstractXRayInterceptor {
     protected void xrayTracedClasses() {
     }
 
-    protected void springRepositories() {
-    }
-
-    /**
-     * @param pjp the proceeding join point
-     * @return the result of the method being wrapped
-     * @throws Throwable
-     */
-    public Object traceAroundRepositoryMethods(ProceedingJoinPoint pjp) throws Throwable {
-        return XRayInterceptorUtils.conditionalProceed(pjp);
-    }
-
     protected Map<String, Map<String, Object>> generateMetadata(ProceedingJoinPoint pjp, Subsegment subsegment) {
         return XRayInterceptorUtils.generateMetadata(pjp, subsegment);
     }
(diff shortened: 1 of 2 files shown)


chanchiem:
Do you have any suggestions on how we can get this tested? 

Can we get some unit tests on this as well?

dekelpilli:
> Do you have any suggestions on how we can get this tested?
> 
> Can we get some unit tests on this as well?

You can test this by simply doing what you would do with a normal Spring Boot app that uses this library, but make sure you don't have the Spring Data JPA library in the classpath. If the behaviour with BaseAbstractXRayInterceptor and no JPA lib is the same as BaseAbstractXRayInterceptor and the JPA lib present, then this change worked.

As for unit tests, I believe this would be non-trivial. Maybe an integration test with some mocked-out AWS-integration parts, but you would definitely need to load the Spring context to get a meaningful test. This is probably the reason that this module currently has no tests at all.

aws-xray-recorder-sdk-spring/src/main/java/com/amazonaws/xray/spring/aop/BaseAbstractXRayInterceptor.java
```diff
@@ -0,0 +1,89 @@
...
+        }
+    }
+
+    protected abstract void xrayEnabledClasses();
+
+    @Pointcut("execution(* XRayTraced+.*(..))")
+    protected void xrayTracedClasses() {
+    }
+
+    protected void springRepositories() {
```
shengxil: There is no need to leave the method in this base class as it intends to be Spring JBI free. We can move this and ```traceAroundRepositoryMethods()``` to the ```AbstractXRayInterceptor```

shengxil:
Thanks for your contribution. I've verified the functionality but left a comment.

aws-xray-recorder-sdk-spring/src/main/java/com/amazonaws/xray/spring/aop/BaseAbstractXRayInterceptor.java
dekelpilli: Good point, done.

PR #128, opened 2020-02-12
Better handling of SQL statements made outside of X-Ray segments.

uldall:
*Description of changes:*
Some applications only require a subset of SQL statements to be tracked by
X-Ray. This change makes such statements not fail with a
SegmentNotFoundException exception.


shengxil:
Hi uldall@,

Thank you for contributing on the X-Ray SDK. By design we throw out the segment not found error when segment is missing to warn the developers that their instrumentation is not working properly instead of just silent failing. If the error is omitted, the subsegment is actually not recorded and the developer may be not aware of it, or the reason. For such situation when the developer really want to ignore the error, we suggest use `LogErrorContextMissingStrategy` by specifying on `AWSXRayRecorder`, or via [environment variable / system property](https://docs.aws.amazon.com/xray/latest/devguide/xray-sdk-java-configuration.html#xray-sdk-java-configuration-envvars).

uldall:
Thank you for the quick review @shengxil. That makes total sense. Closing this PR.

Issue #136, opened 2020-04-01
com.amazonaws.xray.sql.TracingStatement does not handle exceptions correctly.


chamsco-aws:
In the class `com.amazonaws.xray.sql.TracingStatement` if you do not have a current segment and your  `contextMissingStrategy` is set to something other than runtime error then in the following code in the section labeled `// don't trace if failed to create subsegment ` gets ran.

https://github.com/aws/aws-xray-sdk-java/blob/ec0c8ea869df3a5eb7716210a61ee6174b128b87/aws-xray-recorder-sdk-sql/src/main/java/com/amazonaws/xray/sql/TracingStatement.java#L98-L129

The issue is it does not have the same exception handling as the lower try/catch block with the comment `// the reflection may wrap the actual error with an InvocationTargetException...`, the first `if` block never tries to do `throw ite.getTargetException()` or `throw ite.getCause()` this causes a InvocationTargetException to be raised on a error.

The non-tracing path should have the same error unwrapping as the tracing path.

shengxil:
Fix merged, closing

PR #137, opened 2020-04-02
Fixed implementation of TracingStatement when no segment is present.

chamsco-aws:
Solves issues brought up in #136

Adds transparent exception handling from TracingStatement when no segment is present.

Adds new tests to test for both checked and unchecked exceptions also tests the exceptions in a environment with no segment created and `IgnoreErrorContextMissingStrategy` set.

### DIFF
--- a/aws-xray-recorder-sdk-sql/src/main/java/com/amazonaws/xray/sql/TracingStatement.java
+++ b/aws-xray-recorder-sdk-sql/src/main/java/com/amazonaws/xray/sql/TracingStatement.java
@@ -97,8 +97,24 @@ public class TracingStatement {
 
             Subsegment subsegment = createSubsegment();
             if (subsegment == null) {
-                // don't trace if failed to create subsegment
-                return method.invoke(delegate, args);
+                try {
+                    // don't trace if failed to create subsegment
+                    return method.invoke(delegate, args);
+                } catch (Throwable t) {
+                    if (t instanceof InvocationTargetException) {
+                        // the reflection may wrap the actual error with an InvocationTargetException.
+                        // we want to use the root cause to make the instrumentation seamless
+                        InvocationTargetException ite = (InvocationTargetException) t;
+                        if (ite.getTargetException() != null) {
+                            throw ite.getTargetException();
+                        }
+                        if (ite.getCause() != null) {
+                            throw ite.getCause();
+                        }
+                    }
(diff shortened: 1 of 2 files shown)


aws-xray-recorder-sdk-sql/src/main/java/com/amazonaws/xray/sql/TracingStatement.java
```diff
@@ -97,8 +97,24 @@ public Object invoke(Object proxy, Method method, Object[] args) throws Throwabl
 
             Subsegment subsegment = createSubsegment();
             if (subsegment == null) {
-                // don't trace if failed to create subsegment
-                return method.invoke(delegate, args);
+                try {
```
shengxil, line 100: This is quite a duplicated code. I'd recommend refactoring it to
```
           try {
                // execute the query 
                return method.invoke(delegate, args);
            } catch (Throwable t) {
                Throwable rootThrowable = t;
                if (t instanceof InvocationTargetException) {
                    // the reflection may wrap the actual error with an InvocationTargetException.
                    // we want to use the root cause to make the instrumentation seamless
                    InvocationTargetException ite = (InvocationTargetException) t;
                    if (ite.getTargetException() != null) {
                        rootThrowable = ite.getTargetException();
                    } else if (ite.getCause() != null) {
                        rootThrowable = ite.getCause();
                    }
                   ...
          
            if(subsegment != null) {
                 subsegment.addException(rootThrowable);
            }
            throw rootThrowable;
 ...
```. 
and merge with the code below

File: aws-xray-recorder-sdk-sql/src/main/java/com/amazonaws/xray/sql/TracingStatement.java [hunk 2]
```diff
@@ -90,42 +90,39 @@ public static CallableStatement decorateCallableStatement(CallableStatement stat
...
 
-                subsegment.addException(t);
-                throw t;
+                if (subsegment != null) {
+                    subsegment.addException(rootThrowable);
+                }
+                throw rootThrowable;
             } finally {
-                AWSXRay.endSubsegment();
+                if (isExecution(method)) {
```
shengxil, line 123: we should check `subsegment == null` as well because subsegment creation may fail. The previous code failed fast if subsegment creation fails

Issue #92, opened 2019-08-19
v2 libraries require AWS SDK v1.11 to run


ashleymercer:
Since https://github.com/aws/aws-xray-sdk-java/issues/43 this library has provided compatibility with v2 of the standard AWS SDK. However, there are still a large number of dependencies on the (old) v1.11 SDK which means that it's necessary to have *both* sets of libraries on the classpath for this to work correctly:

- potential naming conflicts / code errors arise because a lot of classes are identically named (although in a different package structure)

- increases the bulk of applications since we need to bundle jars for both v1.11 and v2 SDKs

- additional complications such as having e.g. two sets of connection pools (one maintained by each SDK) which can lead to obscure bugs

As I see it, this is potentially quite a big undertaking: it would require rewriting large parts of the `core` package in particular to work exclusively against the v2 SDK.

chanchiem:
Hey,

Thank you very much for bringing this to our attention. As of right now, this dependency is required because the X-Ray SDK uses version 1.11 of the AWS SDK to generate a client used for our rule and target poller threads. We aim to have the dependency on v1.11 of the SDK by the next major version release of our Java SDK. Please stay tuned.

Thanks!

annedroiid:
Are there any further updates on this or timelines for when it might be released?

willarmiros:
Hi @annedroiid,

We have rewritten the core X-Ray SDK to remove any dependencies on AWS SDK v1 in #145. However, we exposed references to AWS SDK classes in some of our public APIs, and as such we must maintain the dependency until our next major version, since removing it will constitute a breaking change.

We can't provide an official timeline on a 4.x release, but the dependency will be removed then.

humanzz:
Hi @willarmiros,
I hope u meant 3.x (not 4.x)? 😄

willarmiros:
@humanzz I sure did! Thanks for catching that.

baldram:
> We have rewritten the core X-Ray SDK to remove any dependencies on AWS SDK v1 ... our next major version, since removing it will constitute a breaking change.

Do you have any advice on how to implement generating custom subsegments using only SDK v2? I mean, by adding a bit more library code to one's project, to then use it in their code, before AWS SDK V3 is released?
Perhaps this could help in eliminating the dependency on AWS SDK V1, in order to avoid exposure to vulnerabilities present in SDK V1.

[...]

PR #145, opened 2020-05-14
Replace calls to X-Ray daemon for rules / targets with a simple JDK-b…

anuraaga:
…ased client to allow removing SDK dependency in the future.

*Issue #, if available:*
For #92

*Description of changes:*

Instead of using the SDK to send X-Ray API requests for rules / targets, we just use the JDK's `HttpUrlConnection`. This is fine because these requests are proxied by the daemon and don't need to be authenticated.

I haven't removed the dependency because we expose the SDK classes in many public APIs. It seems that we have too many public methods that are only for internal use, let's aim to reduce this surface for a 3.0 release, where we should also remove the SDK dependency. When doing so, we can just vendor-in simple models corresponding to the four classes we use.

 I have tried to clean stuff up in the classes I touched, especially missing `final` on many variables and an overuse of powermock (well, technically any use is overuse ;) ). I also picked wiremock as my testing framework since I saw the SDK use it too. Also took the liberty of adding assertj which SDK also uses since exception assertions are much nicer with it.

### DIFF
--- a/aws-xray-recorder-sdk-core/src/test/java/com/amazonaws/xray/strategy/sampling/pollers/TargetPollerTest.java
+++ b/aws-xray-recorder-sdk-core/src/test/java/com/amazonaws/xray/strategy/sampling/pollers/TargetPollerTest.java
@@ -1,11 +1,8 @@
 package com.amazonaws.xray.strategy.sampling.pollers;
 
 import static org.assertj.core.api.Assertions.assertThat;
-import static org.mockito.ArgumentMatchers.any;
-import static org.mockito.Mockito.when;
 
 import java.time.Clock;
-import java.util.Collections;
 
 import org.junit.Rule;
 import org.junit.Test;
@@ -13,7 +10,6 @@ import org.mockito.Mock;
 import org.mockito.junit.MockitoJUnit;
 import org.mockito.junit.MockitoRule;
 
-import com.amazonaws.services.xray.model.SamplingStatisticsDocument;
 import com.amazonaws.xray.internal.UnsignedXrayClient;
 import com.amazonaws.xray.strategy.sampling.manifest.CentralizedManifest;
 
--- a/aws-xray-recorder-sdk-core/src/test/java/com/amazonaws/xray/strategy/sampling/pollers/TargetPollerTest.java
+++ b/aws-xray-recorder-sdk-core/src/test/java/com/amazonaws/xray/strategy/sampling/pollers/TargetPollerTest.java
@@ -1,32 +1,32 @@
 package com.amazonaws.xray.strategy.sampling.pollers;
 
-import com.amazonaws.services.xray.AWSXRayClient;
+import com.amazonaws.xray.internal.UnsignedXrayClient;
 import com.amazonaws.xray.strategy.sampling.manifest.CentralizedManifest;
-import org.junit.Test;
-import org.junit.Assert;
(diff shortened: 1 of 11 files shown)


aws-xray-recorder-sdk-core/src/test/java/com/amazonaws/xray/strategy/sampling/pollers/TargetPollerTest.java
```diff
@@ -1,32 +1,32 @@
...
-import org.powermock.reflect.Whitebox;
+import org.mockito.Mock;
+import org.mockito.junit.MockitoJUnit;
+import org.mockito.junit.MockitoRule;
 
-import java.time.Clock;
-import java.util.concurrent.ScheduledExecutorService;
+import static org.junit.Assert.assertTrue;
 
 public class TargetPollerTest {
```
willarmiros, line 14: Can we add a `testPollerNotAbruptMainThread` for the target poller as well, for parity?

anuraaga: I did the opposite, removing the test from `RulePoller`. I don't think these actually test anything. I noticed this comment

https://github.com/aws/aws-xray-sdk-java/blob/master/aws-xray-recorder-sdk-core/src/main/java/com/amazonaws/xray/strategy/sampling/pollers/TargetPoller.java#L43

`// An Error should not be handled by the application. The executor will die and not abrupt main thread.`

This comment doesn't make much sense to me - I'm not aware of any form of business logic that would cause the main thread to be interrupted from an executor (we're not passing the result back using e.g. `Future`) so I don't think it's possible for these tests to ever fail no matter what the code is. So they don't appear to test anything.

I don't know about executors dying either, even if a callback throws, the executor should still continue just fine.

Let me know if I'm missing anything.

shengxil: The test is indeed an invalid one. Executor would never abrupt the main thread.

I think the comment really should be
"The executor omits any exception except `Error`, which should stop the executor."

`ScheduledExecutor` dies if there's an unhandled exception. The below snippet will stop printing msg at the 5th execution
```
        AtomicInteger i = new AtomicInteger();
        Executors.newSingleThreadScheduledExecutor().scheduleAtFixedRate(() -> {
            if(i.incrementAndGet() == 5) {
                throw new IllegalStateException("STOP");
            }
            System.out.println("Hello World");
        }, 0, 1000, TimeUnit.MILLISECONDS);
```

anuraaga: Oops - thanks for clarifying the `ExecutorService` behavior, never noticed that!

Tweaked the wording of the comment. Believe the behavior should be tweaked a bit to follow how I've seen most libraries propagate fatal errors (`AssertionError` is also `Error` but definitely not fatal...) but will look at that in another PR.

willarmiros: Good call @shengxil, LGTM @anuraaga

Issue #151, opened 2020-05-22
Errors connected to xray instrumentation should not break business processing


kwalczak:
In current implementation, not all exceptions are caught in `com.amazonaws.xray.emitters.UDPEmitter#sendData` resulting in exceptions bubbling up.

Eg. restart/crash of xray agent should result only in log in application - not exceptions.

anuraaga:
Fixed by #152

PR #152, opened 2020-05-22
Extended try block in UDPEmitter (#151)

kwalczak:
(#151)

- fixed bubbling up exceptions thrown during sending segment to xray daemon

### DIFF
--- a/aws-xray-recorder-sdk-core/src/test/java/com/amazonaws/xray/emitters/UDPEmitterTest.java
+++ b/aws-xray-recorder-sdk-core/src/test/java/com/amazonaws/xray/emitters/UDPEmitterTest.java
@@ -1,15 +1,12 @@
 package com.amazonaws.xray.emitters;
 
+import static com.amazonaws.xray.AWSXRay.getGlobalRecorder;
+import static org.assertj.core.api.Assertions.assertThat;
+
 import com.amazonaws.xray.config.DaemonConfiguration;
 import com.amazonaws.xray.entities.DummySegment;
-import org.junit.Test;
-
-import java.net.DatagramSocket;
 import java.net.SocketException;
-
-import static com.amazonaws.xray.AWSXRay.getGlobalRecorder;
-import static org.junit.Assert.assertEquals;
-import static org.mockito.Mockito.mock;
+import org.junit.Test;
 
 public class UDPEmitterTest {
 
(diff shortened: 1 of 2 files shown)


aws-xray-recorder-sdk-core/src/test/java/com/amazonaws/xray/emitters/UDPEmitterTest.java
```diff
@@ -1,21 +1,44 @@
...
+    @Test
+    public void sendingSegmentShouldNotThrowExceptions() throws SocketException {
+        DaemonConfiguration config = getDaemonConfiguration("host:1234");
+        UDPEmitter emitter = new UDPEmitter(config);
+
+        try {
+            final boolean success = emitter.sendSegment(new DummySegment(getGlobalRecorder()));
+            assertEquals(false, success);
+        } catch (Exception e) {
+            fail();
```
anuraaga, line 35: An exception will cause the test to fail anyways so we don't need to catch, can you remove the try clause?

aws-xray-recorder-sdk-core/src/test/java/com/amazonaws/xray/emitters/UDPEmitterTest.java
```diff
@@ -1,21 +1,44 @@
...
+    }
+
+
+    @Test
+    public void sendingSegmentShouldNotThrowExceptions() throws SocketException {
+        DaemonConfiguration config = getDaemonConfiguration("host:1234");
+        UDPEmitter emitter = new UDPEmitter(config);
+
+        try {
+            final boolean success = emitter.sendSegment(new DummySegment(getGlobalRecorder()));
```
willarmiros, line 32: Could you mock the UDPEmitter's datagram socket so that this unit test does not actually use the network? Adding a new package-private constructor or datagram socket setter is fine.

kwalczak:
> Thanks for the patch! Just for reference what exception were you seeing?

```
java.lang.IllegalArgumentException: unresolved address
	at java.base/java.net.DatagramPacket.setSocketAddress(DatagramPacket.java:316)
	at java.base/java.net.DatagramPacket.<init>(DatagramPacket.java:144)
	at java.base/java.net.DatagramPacket.<init>(DatagramPacket.java:178)
	at com.amazonaws.xray.emitters.UDPEmitter.sendData(UDPEmitter.java:87)
	at com.amazonaws.xray.emitters.UDPEmitter.sendSegment(UDPEmitter.java:70)
	at com.amazonaws.xray.AWSXRayRecorder.sendSegment(AWSXRayRecorder.java:159)
	at com.amazonaws.xray.AWSXRayRecorder.endSegment(AWSXRayRecorder.java:442)
	at com.amazonaws.xray.javax.servlet.AWSXRayServletFilter.postFilter(AWSXRayServletFilter.java:395)
	at com.amazonaws.xray.javax.servlet.AWSXRayServletFilter.doFilter(AWSXRayServletFilter.java:169)
```

aws-xray-recorder-sdk-core/src/test/java/com/amazonaws/xray/emitters/UDPEmitterTest.java
anuraaga: Unfortunately from the stacktrace we can see it's the `DatagramPacket` constructor itself that throws, so it doesn't seem possible to completely avoid the network. Definitely preferable if we could, but for now since it didn't add value I went ahead and reverted the change to add a new constructor, we can bring it back if we find a different use case for it.

willarmiros: Ah I see, thanks for fixing that. Didn't think about how mocking could obfuscate the source of exception.

PR #153, opened 2020-05-25
Add Gradle build configuration.

anuraaga:
Adds a configuration to build the project with Gradle. Gradle usually runs faster especially for large projects, avoiding rerunning the same tasks / compilation multiple times when files haven't changed. I want to add more tooling to this project, checkstyle, errorprone, coverage, etc, and it makes sense to migrate to Gradle before rather than after adding too much tooling.

I have sanity checked the POMs and javadoc they look fine.

Next steps in migrating to Gradle in separate PRs

- Migrate benchmark to Gradle
- Fork / update release documentation
- Verify release works
- Remove Maven configuration / old release documentation


gradle.properties
```diff
@@ -0,0 +1 @@
+version = 2.5.1-SNAPSHOT
```
willarmiros, line 1: Why 2.5.1 instead of 2.5.0? And why SNAPSHOT?

build.gradle.kts
```diff
@@ -0,0 +1,119 @@
...
+                    val folder = file("${propertiesDir}/com/amazonaws/xray")
+                    folder.mkdirs()
+                    val propertiesFile = folder.resolve("sdk.properties")
+                    propertiesFile.writeText("awsxrayrecordersdk.version=${project.version}")
+                }
+            }
+        }
+    }
+
+    plugins.withId("maven-publish") {
```
willarmiros, line 64: Where is GPG signing done? Do we need to configure the target repo for publishing like we do [here](https://github.com/aws/aws-xray-sdk-java/blob/master/pom.xml#L74-L80)?

gradle.properties
anuraaga: It's conventional for Java projects to switch to a snapshot version number after releasing, and the main reason I went ahead and did it is because it allows `mavenLocal` to have less priority than `mavenCentral` (may notice that in the `repositories` block in the build.gradle.kts file). I've had many instances of confusion with the other order and having a local version of an artifact override a released one which should basically never be done.

build.gradle.kts
anuraaga: Good eye - I was thinking of following up with verifying the config and commands for releasing but at least adding the publishing repo now is straightforward so added it.

willarmiros: Fair enough, since we skip signing for non-release builds anyway. We'll come back to this.

gradle.properties
willarmiros: Ok makes sense, but still have a couple questions:

* If the main reason for adding the SNAPSHOT is to distinguish it from the version already released in `mavenCentral`, why do we also increment the patch version number? Why not just make it 2.5.0-SNAPSHOT?
* If we want to depend on the local version of the repo, say to test local-only changes with a sample app, we would depend on the SNAPSHOT version right?
* When making the next release, say 2.6.0, would we remove the SNAPSHOT suffix before releasing and then replace it afterward?

build.gradle.kts
willarmiros: For reference, I remember there was some issues about signing in Gradle with the X-Ray OT SDK: https://github.com/awslabs/aws-xray-sdk-with-opentelemetry/blob/master/build.gradle#L80-L89

But as per the linked GitHub issue, it looks to be resolved with Gradle 6.2 so we should be good.

anuraaga: Thanks for this pointer :)

gradle.properties
anuraaga: - Yup, the snapshot is not production and just to provide previews, as long as the number is > than the previous released version and <= the next release, it works out as an interim build. So it's simple to just increment the patch number for SNAPSHOT even if the release ends up being a minor

- Yeah would depend on the SNAPSHOT version, and I believe for local no need to worry about update policy, that's for fetching snapshots from remote

- Yeah most libraries use maven release plugin or gradle release plugin which automates that. Will see if it works with our release process, and if for some reason it doesn't I'll revert to just using a normal version number :)

PR #156, opened 2020-06-01
Apply checkstyle and prepare for license header check.

anuraaga:
This applies Gradle plugins to apply license header and checkstyle. License header and IntelliJ / checkstyle config are copied from the AWS SDK.

Most code changes are license addition and whitespace. Sometimes, I cleaned up code that otherwise would cause checkstyle to fail, e.g.,

- `private ObjectMapper mapper = ...` -> `private static final ObjectMapper MAPPER = `
- Deprecate utility classes without a private constructor. While it's really unlikely that users call them, it's not impossible so we should fix in a major release. As these utilities are for the SDK they should be internal anyways so just went ahead and deprecated them. We'll probably deprecate a lot more stuff that users probably don't use in an effort to reduce surface for v3.


anuraaga:
Thanks - while waiting on the license plugin, I went ahead and disabled it for now. Enabling license is quite simple and automated while checkstyle is much harder so will go ahead and merge just that in for now to prevent drift.

PR #167, opened 2020-06-11
Depend on aws sdk modules from instrumentor for runtime rather than c…

anuraaga:
…ompile only.

The dependencies are needed when a user app runs too, so `compileOnly`, which indicates it's only needed when this module is compiled, wasn't right and would require users to add both dependencies.

Note to prevent confusion as it's happening right now, this does not affect our next release, which we're still using Maven for. It probably would affect the one after that where we should have switched to Gradle by then.

### DIFF
--- a/aws-xray-recorder-sdk-aws-sdk-instrumentor/build.gradle.kts
+++ b/aws-xray-recorder-sdk-aws-sdk-instrumentor/build.gradle.kts
@@ -4,7 +4,7 @@ plugins {
 }
 
 dependencies {
-    compileOnly(project(":aws-xray-recorder-sdk-aws-sdk"))
+    implementation(project(":aws-xray-recorder-sdk-aws-sdk"))
 }
 
 description = "AWS X-Ray Recorder SDK for Java - AWS SDK Instrumentor"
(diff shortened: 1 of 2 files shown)


aws-xray-recorder-sdk-aws-sdk-instrumentor/build.gradle.kts
```diff
@@ -4,7 +4,7 @@ plugins {
 }
 
 dependencies {
-    compileOnly(project(":aws-xray-recorder-sdk-aws-sdk"))
+    implementation(project(":aws-xray-recorder-sdk-aws-sdk"))
```
willarmiros, line 7: We actually [don't have a dependency](https://github.com/aws/aws-xray-sdk-java/blob/master/aws-xray-recorder-sdk-aws-sdk-instrumentor/pom.xml) on the `aws-xray-recorder-sdk-aws-sdk` module currently. But that should be fine since the customers who are depending on both right now would just have it deduped. We should patch the [docs](https://docs.aws.amazon.com/xray/latest/devguide/xray-sdk-java-awssdkclients.html) when this is released.

PR #173, opened 2020-06-22
Add GitHub Actions build.

anuraaga:
*Description of changes:*

GitHub actions are well integrated with the GitHub UI and make it easy to verify this builds on multiple OSes. Let's migrate off of Travis.

This only adds the Gradle build since we'll remove the Maven build soon I think.

Also enabled parallel build for Gradle.

### DIFF
--- a/.github/workflows/continuous-build.yml
+++ b/.github/workflows/continuous-build.yml
@@ -1,0 +1,37 @@
+name: Continuous Build
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - master
+
+jobs:
+  build:
+    name: Build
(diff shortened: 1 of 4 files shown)


anuraaga:
https://github.com/aws/aws-xray-sdk-java/pull/173/commits/e8008282ae8bd0b9365d41db73c2b1bd3b910b3f fixes an incorrect test assertion, it should be looking at the raw address string but is instead triggering a reverse DNS lookup. This has behavior that depends on the host, and in particular does not pass on GitHub's Windows CI.

For reference, this is the run against my branch https://github.com/anuraaga/aws-xray-sdk-java/runs/794461143?check_suite_focus=true

PR #176, opened 2020-07-09
Remove unused POM file.

anuraaga:
Happy to close this PR if this file is actually needed. I can't find any reason for it. I notice similar content is in the documentation

https://docs.aws.amazon.com/xray/latest/devguide/xray-sdk-java.html

But that's still showing 2.4.0 so assuming there is no automatic sync. I'd rather replace the example in our README with the BOM version for both maven and Gradle. 


willarmiros:
@shengxil would be more likely to know the reason, if any, behind this POM.

shengxil:
Confirming the POM is a stale file. Approving

PR #177, opened 2020-07-09
Use dynamic version based on previous git tag and add task for releas…

anuraaga:
…ing.

This sets the version of the project automatically based on latest tag

- Normal: Increment minor version and suffix with -SNAPSHOT
- With `-Preckon.stage=final` parameter: Increment minor version

The `release` tag fills the version into `README.md`, commits it, and builds / publishes. In a followup I will wire up the benchmarks as well.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -7,6 +7,28 @@ plugins {
     id("com.github.hierynomus.license") apply false
     id("net.ltgt.errorprone") apply false
     id("org.checkerframework") apply false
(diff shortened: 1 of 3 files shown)


willarmiros:
One question - if we want to bump something other than the minor version, how do we accomplish that?

anuraaga:
Great question, there's a property to control that so we'd pass `-Preckon.scope=patch` when releasing

PR #180, opened 2020-07-14
Return no-op when segment context isn't found instead of null.

anuraaga:
While in the strictest sense, changing behavior like this is a breaking change, it's generally acceptable to replace a null with a no-op (but not vice versa). The idea is that returning the no-op segment should work in any caller code in the non-null case no problem. On the flip side, there should be less chance of NPE in user code as a result - all the major Java tracing libraries make sure never to return null spans so it's fairly standard practice.

PS: I'm intentionally not using the name of the actual class here, it's something we'll need to rename for the next major version.


aws-xray-recorder-sdk-core/src/test/java/com/amazonaws/xray/AWSXRayRecorderTest.java
```diff
@@ -443,83 +447,83 @@ public void testEmittingSegmentTwiceThrowsSegmentAlreadyEmittedException() {
...
+    @Test
+    public void testSubsegmentFunctionExceptionWhenMissingContext() {
+        IllegalStateException expectedException = new IllegalStateException("To be thrown by function");
         Function<Subsegment, Void> function = (subsegment) -> {
             throw expectedException;
         };
+
         AWSXRayRecorder recorder = AWSXRayRecorderBuilder.standard()
-                                                         .withContextMissingStrategy(new LogErrorContextMissingStrategy())
+                                                         .withContextMissingStrategy(new IgnoreErrorContextMissingStrategy())
```
anuraaga, line 484: These tests don't actually test whether anything's logged so I just removed that to make them more precise.

aws-xray-recorder-sdk-core/src/main/java/com/amazonaws/xray/javax/servlet/AWSXRayServletFilter.java
```diff
@@ -352,7 +345,7 @@ public Segment preFilter(ServletRequest request, ServletResponse response) {
         final Segment created;
         if (SampleDecision.SAMPLED.equals(sampleDecision)) {
             created = recorder.beginSegment(getSegmentName(httpServletRequest), traceId, parentId);
-            if (created != null && samplingResponse.getRuleName().isPresent()) {
+            if (samplingResponse.getRuleName().isPresent()) {
                 logger.debug("Sampling strategy decided to use rule named: " + samplingResponse.getRuleName().get() + ".");
                 created.setRuleName(samplingResponse.getRuleName().get());
             }
```
willarmiros, line 351: Think you can get rid of the other null check in the `else` block

Issue #28, opened 2018-09-16
SQL tracing: sql does not show in console


grobmeier:
I am running a Spring webapp with this configuration:

dataSource.setJdbcInterceptors("com.amazonaws.xray.sql.mysql.TracingInterceptor;");

Debugging confirms I am running into the TracingInterceptor and inside that, I can even see the sql inside the TracingInterceptor. 

However, in the AWS XRay Console I cannot see the SQL commandos fired, just some meta data (like connection information).

Am I supposed to see the raw SQL running against a database? If yes, is there any additional step I missed?

svetozar02:
Any updates on this issue? 
From my perspective, there isn't much value in tracing sql queries if you don't see the actual sql statement. At very least it should be mentioned at https://docs.aws.amazon.com/xray/latest/devguide/xray-sdk-java-sqlclients.html so users know what to expect.

willarmiros:
Hi @svetozar02,
Unfortunately it is still a security concern to capture arbitrary SQL query information in subsegments, and we cannot add this information at this time. I've put in a request to explicitly mention this on the documentation you linked.

svetozar02:
Would it be possible for have an option of enabling it if users actually want it?

Other option is to look at how Datadog agent is obfuscating sql...

softprops:
if not automatic, is there any guidance on a hook consumers can do in code to provide the instrumentation the sql manually?

softprops:
The docs for the node sdk provided a tip that also works for the Java sdk https://docs.aws.amazon.com/xray/latest/devguide/xray-sdk-nodejs-sqlclients.html

willarmiros:
Hi all,

Sorry for the delay in response here. You can capture SQL queries automatically with the [X-Ray Java Agent](https://docs.aws.amazon.com/xray/latest/devguide/aws-x-ray-auto-instrumentation-agent-for-java.html). You just have to opt-in by setting the `collectSqlQueries` parameter in the agent config file to true. 

In the meantime, this small feature is in our backlog to add to the Java SDK, and we are also always open to pull requests. We need to make this opt in via an environment variable like `AWS_XRAY_COLLECT_SQL_QUERIES`.

PR #186, opened 2020-07-27
Refactored createSqlSubsegment into a public util file

willarmiros:
*Description of changes:*
In order to reuse the subsegment creation logic in the Agent's SQL auto-instrumentation, I moved the private `createSubsegment` method into a new public util class. I made the new Util class rather than just making the method public in `TracingStatement` since we've deprecated TracingStatement, so I thought this public method should be separate.

I also added a capability to record SQL queries, relating to #28. Including them in subsegments will be configurable for the agent, but no such config exists in the SDK yet.

### DIFF
--- a/aws-xray-recorder-sdk-sql/src/test/java/com/amazonaws/xray/sql/SqlSubsegmentsTest.java
+++ b/aws-xray-recorder-sdk-sql/src/test/java/com/amazonaws/xray/sql/SqlSubsegmentsTest.java
@@ -15,7 +15,7 @@
 
 package com.amazonaws.xray.sql;
 
-import static org.junit.Assert.assertEquals;
+import static org.assertj.core.api.Assertions.assertThat;
 import static org.mockito.Mockito.when;
 
 import com.amazonaws.xray.AWSXRay;
@@ -30,13 +30,13 @@ import java.util.Map;
 
 import org.junit.After;
 import org.junit.Before;
+import org.junit.Rule;
 import org.junit.Test;
-import org.junit.runner.RunWith;
 import org.mockito.Mock;
-import org.mockito.runners.MockitoJUnitRunner;
+import org.mockito.junit.MockitoJUnit;
+import org.mockito.junit.MockitoRule;
 
-@RunWith(MockitoJUnitRunner.class)
-public class SqlUtilsTest {
+public class SqlSubsegmentsTest {
     private static final String URL = "http://www.foo.com";
     private static final String HOST = "www.foo.com";
     private static final String USER = "user";
@@ -46,6 +46,9 @@ public class SqlUtilsTest {
     private static final String SQL = "sql";
     private static final String CATALOG = "catalog";
     private Map<String, Object> expectedSqlParams;
+
+    @Rule
+    public MockitoRule rule = MockitoJUnit.rule();
 
     @Mock
     Connection connection;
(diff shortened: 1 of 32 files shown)


aws-xray-recorder-sdk-sql/src/test/java/com/amazonaws/xray/sql/SqlUtilsTest.java
```diff
@@ -0,0 +1,99 @@
...
+    @Test
+    public void testCreateSubsegmentWithoutSql() throws SQLException {
+        Subsegment sub = SqlUtils.createSqlSubsegment(connection, null);
+        verifySqlSubsegment(sub);
+    }
+
+    @Test
+    public void testCreateSubsegmentWithSql() throws SQLException {
+        Subsegment sub = SqlUtils.createSqlSubsegment(connection, SQL);
+        expectedSqlParams.put("sanitized_query", SQL);
```
anuraaga, line 90: It's hard to understand the expectation when modifying a field like this, copy-paste tends to make for clearer tests, a helper may be useful for an extremely huge amount of assertion code though, doesn't look that helpful here.

aws-xray-recorder-sdk-sql/src/test/java/com/amazonaws/xray/sql/SqlUtilsTest.java
```diff
@@ -0,0 +1,99 @@
...
+        verifySqlSubsegment(sub);
+    }
+
+    private void verifySqlSubsegment(Subsegment sub) {
+        assertEquals(CATALOG + "@" + HOST, sub.getName());
```
anuraaga, line 95: For new tests it'd be nice if we can use assertj. For the map, assertj's `assertThat(sub.getSql()).containsExactly` for example

aws-xray-recorder-sdk-sql/src/test/java/com/amazonaws/xray/sql/SqlUtilsTest.java
```diff
@@ -0,0 +1,99 @@
...
+import java.util.Map;
+
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mockito.Mock;
+import org.mockito.runners.MockitoJUnitRunner;
+
+@RunWith(MockitoJUnitRunner.class)
```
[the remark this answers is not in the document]
willarmiros: I tried to do this but kept getting NPEs where I was stubbing methods because the mocks weren't initialized. I tried to also use `ExtendWith` like you did in some other classes, but that caused a weird `NoSuchMethodError`. I ended up settling on the static `initMocks` method in the `BeforeEach` block.

aws-xray-recorder-sdk-sql/src/test/java/com/amazonaws/xray/sql/SqlSubsegmentsTest.java
```diff
@@ -0,0 +1,105 @@
...
+
+    @Mock
+    Connection connection;
+
+    @Mock
+    DatabaseMetaData metaData;
+
+    @BeforeEach
+    void setup() throws SQLException {
+        MockitoAnnotations.initMocks(this);
```
anuraaga, line 56: Hmm surprised extends with didn't work will take a look after the PR

PR #189, opened 2020-08-04
made SqlSubsegments handle exceptions internally

willarmiros:
*Description of changes:*
Follow up to #186 after actually using the change, I realized that there could be some weird inconsistent state behavior the way I implemented it. An unchecked `SQLException` could be thrown both before and after the subsegment was begun, so it was not a guarantee that calling the method would always create a subsegment. This was a recipe for bad behavior/segment leaks. 

Now, we catch all the SQL Exceptions, log useful messages, and make absolutely sure *some* sort of subsegment is began since that is the expected behavior.

Also exposed some of the SQL subsegment fields, since they are the publicly documented whitelisted values (e.g. won't be changing often).

### DIFF
--- a/aws-xray-recorder-sdk-sql/src/main/java/com/amazonaws/xray/sql/SqlSubsegments.java
+++ b/aws-xray-recorder-sdk-sql/src/main/java/com/amazonaws/xray/sql/SqlSubsegments.java
@@ -36,43 +36,55 @@ public final class SqlSubsegments {
     private static final Log logger = LogFactory.getLog(SqlSubsegments.class);
 
     // https://docs.aws.amazon.com/xray/latest/devguide/xray-api-segmentdocuments.html#api-segmentdocuments-sql
-    private static final String DEFAULT_DATABASE_NAME = "database";
     private static final String URL = "url";
     private static final String USER = "user";
-    private static final String DRIVER_VERSION = "driver_version";
-    private static final String DATABASE_TYPE = "database_type";
-    private static final String DATABASE_VERSION = "database_version";
-    private static final String SANITIZED_QUERY = "sanitized_query";
+
+    public static final String DRIVER_VERSION = "driver_version";
+    public static final String DATABASE_TYPE = "database_type";
+    public static final String DATABASE_VERSION = "database_version";
+    public static final String SANITIZED_QUERY = "sanitized_query";
+    public static final String DEFAULT_DATABASE_NAME = "database";
 
     /**
      * Begins a {@link Subsegment} populated with data provided by the {@link Connection#getMetaData} method. Includes
-     * the SQL query string if it is non-null, omits it otherwise.
+     * the SQL query string if it is non-null, omits it otherwise. Takes care to swallow any potential
+     * {@link SQLException}s and always start a subsegment for consistency.
      *
      * @param connection the JDBC connection object used for the query this {@link Subsegment} represents.
      * @param query the SQL query string used in this query, or {@code null} if it is not desirable to include in the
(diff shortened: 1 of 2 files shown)


aws-xray-recorder-sdk-sql/src/main/java/com/amazonaws/xray/sql/SqlSubsegments.java
```diff
@@ -36,43 +36,55 @@
...
     // https://docs.aws.amazon.com/xray/latest/devguide/xray-api-segmentdocuments.html#api-segmentdocuments-sql
-    private static final String DEFAULT_DATABASE_NAME = "database";
     private static final String URL = "url";
     private static final String USER = "user";
-    private static final String DRIVER_VERSION = "driver_version";
-    private static final String DATABASE_TYPE = "database_type";
-    private static final String DATABASE_VERSION = "database_version";
-    private static final String SANITIZED_QUERY = "sanitized_query";
+
+    public static final String DRIVER_VERSION = "driver_version";
```
anuraaga, line 42: Can we add javadoc for these new public fields?

aws-xray-recorder-sdk-sql/src/test/java/com/amazonaws/xray/sql/SqlSubsegmentsTest.java
```diff
@@ -102,4 +102,15 @@ void testCreateSubsegmentWithSql() throws SQLException {
...
+
+    @Test
+    void testCreateSubsegmentWhenConnectionThrowsException() throws SQLException {
+        when(connection.getMetaData()).thenThrow(new SQLException());
+
+        Subsegment sub = SqlSubsegments.forQuery(connection, SQL);
+
+        assertThat(AWSXRay.getCurrentSubsegment()).isEqualTo(sub);
+        assertThat(sub.isInProgress()).isTrue();
+        assertThat(sub.getParentSegment().getSubsegments()).containsOnly(sub);
```
anuraaga, line 114: I think the intent of this check is captured better as just `contains`

anuraaga: Think we can use an assertion on the subsegment name, both as a check and documentation of what happens in this case.

aws-xray-recorder-sdk-sql/src/main/java/com/amazonaws/xray/sql/SqlSubsegments.java
```diff
@@ -36,43 +36,55 @@
...
-
         subsegment.setNamespace(Namespace.REMOTE.toString());
-        subsegment.putSql(URL, metadata.getURL());
-        subsegment.putSql(USER, metadata.getUserName());
-        subsegment.putSql(DRIVER_VERSION, metadata.getDriverVersion());
-        subsegment.putSql(DATABASE_TYPE, metadata.getDatabaseProductName());
-        subsegment.putSql(DATABASE_VERSION, metadata.getDatabaseProductVersion());
+
+        try {
+            subsegment.putSql(URL, metadata.getURL());
```
anuraaga, line 80: I don't think we need to assert every one of these but one of them to give some code coverage is still a good idea.

aws-xray-recorder-sdk-sql/src/test/java/com/amazonaws/xray/sql/SqlSubsegmentsTest.java
willarmiros: Good calls

PR #191, opened 2020-08-07
added stacktrace to Log error strat

willarmiros:
*Description of changes:*
Adds `debug`-level logging of the stacktrace for context missing exceptions that are suppressed by the Log Error strategy. This could be helpful for both customers and us to debug CME issues, especially in the agent whose default strategy is Log Error.


aws-xray-recorder-sdk-core/src/main/java/com/amazonaws/xray/strategy/LogErrorContextMissingStrategy.java
```diff
@@ -32,6 +32,7 @@
     @Override
     public void contextMissing(String message, Class<? extends RuntimeException> exceptionClass) {
         logger.error("Suppressing AWS X-Ray context missing exception (" + exceptionClass.getSimpleName() + "): " + message);
+        logger.debug(new RuntimeException(message));
```
anuraaga, line 35: The exception constructor has a lot of overhead, so we need to call it only when debug is enabled

willarmiros: Thanks, a bit tired :)

Issue #117, opened 2019-12-26
Implementation of log group discovery on EC2 not work on Beanstalk


hugoprudente:
Hi all,

The feature [Implementation of log group discovery on EC2](https://github.com/aws/aws-xray-sdk-java/commit/08629ee2a61ea73d85c6af0843150fd7d9399617) released on the **AWS X-RAY SDK 2.4.0** and not described on the [CHANGELOG](https://github.com/aws/aws-xray-sdk-java/blob/master/CHANGELOG.md) does **not work** properly on Beanstalk.

```java
WARN com.amazonaws.xray.plugins.EC2Plugin.populateLogReferences(EC2Plugin.java:121) []  -  
CloudWatch Agent log configuration file not found at /opt/aws/amazon-cloudwatch-agent/etc/log-config.json. 
Install the CloudWatch Agent on this instance to record log references in X-Ray.

```

The code is searching for the "CloudWatch Logs" configuration on the following path HARDCODED.

https://github.com/aws/aws-xray-sdk-java/blob/c183a3aad20a528a69cff26d77dfb691f503edb9/aws-xray-recorder-sdk-core/src/main/java/com/amazonaws/xray/plugins/EC2Plugin.java#L44

The problem is that the **ElasticBeanstalk** platform uses the does not use the `/opt` as the CloudWatch Logs path, neither for it's configurations:

```
[root@ip-172-31-0-213 ec2-user]# ps aux | grep log
root      2528  0.8  1.2 913352 50876 ?        SNl  09:22   0:13 /usr/bin/python2.7 /usr/bin/aws logs push --config-file /etc/awslogs/awslogs.conf --additional-configs-dir /etc/awslogs/config

[root@ip-172-31-0-213 ec2-user]# tree /opt/aws -L 1
/opt/aws
├── amitools
├── apitools
├── bin
└── credential-file-path.template
```
The above issue also caused the a miss on the feature #110 referred to the code below returning this message:

```java
WARN com.amazonaws.xray.AWSXRayRecorderBuilder.lambda$build$1(AWSXRayRecorderBuilder.java:249) []  -  
com.amazonaws.xray.plugins.EC2Plugin plugin returned empty Log References. 
The recorder will not reflect the logs from this plugin.
WARN com.amazonaws.xray.AWSXRayRecorderBuilder.lambda$build$1(AWSXRayRecorderBuilder.java:249) []  -  
[...]

PR #194, opened 2020-08-17
added lazy initialization to improve withDefaultPlugins performance

willarmiros:
*Issue #, if available:*
#117 

*Description of changes:*
In order to improve the time it takes to quickly construct and check `isEnabled` on all plugins, I adjusted a couple other plugins to fail faster (related to #192):

* In `ElasticBeanstalkPlugin`, made `ObjectMapper` initialization lazy
* In `EKSPlugin`, made call to `getClusterName` lazy

Also reduced logging level for missing log groups to `debug`, since log group discovery is not yet implemented for all plugins and we shouldn't emit a warning in that case.


aws-xray-recorder-sdk-core/src/main/java/com/amazonaws/xray/plugins/ElasticBeanstalkPlugin.java
```diff
@@ -62,6 +64,12 @@ public boolean isEnabled() {
 
     public void populateRuntimeContext() {
         byte[] manifestBytes = new byte[0];
+
+        // lazily initialize for performance on unused plugins
+        if (OBJECT_MAPPER == null) {
+            OBJECT_MAPPER = new ObjectMapper();
```
anuraaga, line 70: Since this is only called once anyways I guess there's no point in even having a `static` variable if we're lazying

anuraaga: Ah sorry I meant we can just have local variable

`ObjectMapper objectMapper = new ObectMapper();`

willarmiros: I see my bad.

PR #198, opened 2020-08-26
Updated Benchmarking

willarmiros:
*Description of changes:*
Unified the benchmarks to have similar configurations, which are:
* Average throughput and sample time modes (as opposed to all modes, which IMO didn't add much value and was a pain to maintain)
* 1 fork (since that's what the `jmh` task was set to)
* 10 warmup iterations, 5 measured iterations, 1 second per iteration
* Microsecond output (since 10^-4 ops/ns isn't very helpful)

Updated the readme instructions to run benchmarks with Gradle instead of Maven. Also updated Gradle publishing to exclude unnecessary `.module` files from being published.

These changes allow us to run benchmarks, including the newer trace ID ones, in 20 minutes as opposed to 70+ minutes that it took before, which will make releases a little easier in the future.

### DIFF
--- a/aws-xray-recorder-sdk-benchmark/README.md
+++ b/aws-xray-recorder-sdk-benchmark/README.md
@@ -5,7 +5,15 @@ Benchmarking the Java SDK is necessary to help isolate performance issues and bo
 To run the benchmark, make sure you have Maven and Java 8 or above installed. With Maven, it will download the necessary X-Ray dependencies that the benchmark uses as well as the JMH core and annotation processors that this benchmark uses.
 
 To run the benchmark, use your favorite shell to go into the directory. Then run the following commands:
-```BASH
+```shell script
+git clone https://github.com/aws/aws-xray-sdk-java.git
+cd aws-xray-sdk-java
+./gradlew jmh
+```
+
+Or, alternatively, you can execute the created JAR directly:
+
+```shell script
 git clone https://github.com/aws/aws-xray-sdk-java.git
 cd aws-xray-sdk-java
 ./gradlew jmhJar
--- a/aws-xray-recorder-sdk-benchmark/README.md
+++ b/aws-xray-recorder-sdk-benchmark/README.md
@@ -8,7 +8,8 @@ To run the benchmark, use your favorite shell to go into the directory. Then run
 ```BASH
 git clone https://github.com/aws/aws-xray-sdk-java.git
 cd aws-xray-sdk-java
-./gradlew jmh
+./gradlew jmhJar
+java -jar aws-xray-recorder-sdk-benchmark/build/libs/aws-xray-recorder-sdk-benchmark-<VERSION>-jmh.jar
 ```
 
 You should then start seeing the output of the benchmark:
@@ -39,9 +40,9 @@ You should then start seeing the output of the benchmark:
 
 If you wish to run a specific benchmark, please run the following command:
 ```
-java -jar target/benchmarks.jar <Benchmark_name_here>
+java -jar aws-xray-recorder-sdk-benchmark-<VERSION>-jmh.jar <Benchmark_name_here>
 
-# Example: java -jar target/benchmarks.jar com.amazonaws.xray.AWSXRayRecorderBenchmark.beginDummySegmentBenchmark -w 1 -r 1
+# Example: java -jar aws-xray-recorder-sdk-benchmark-<VERSION>-jmh.jar com.amazonaws.xray.AWSXRayRecorderBenchmark.beginDummySegmentBenchmark
 ```
 
 ## Benchmark Results
--- a/aws-xray-recorder-sdk-benchmark/README.md
+++ b/aws-xray-recorder-sdk-benchmark/README.md
@@ -6,9 +6,9 @@ To run the benchmark, make sure you have Maven and Java 8 or above installed. Wi
 
 To run the benchmark, use your favorite shell to go into the directory. Then run the following commands:
 ```BASH
-cd aws-xray-sdk-java/aws-xray-recorder-sdk-benchmark/
-mvn clean && mvn package
-java -jar target/benchmark.jar
+git clone https://github.com/aws/aws-xray-sdk-java.git
+cd aws-xray-sdk-java
+./gradlew jmh
 ```
 
 You should then start seeing the output of the benchmark:
--- a/aws-xray-recorder-sdk-benchmark/tst/main/java/com/amazonaws/xray/AWSXRayRecorderBenchmark.java
+++ b/aws-xray-recorder-sdk-benchmark/tst/main/java/com/amazonaws/xray/AWSXRayRecorderBenchmark.java
@@ -56,16 +56,6 @@ public class AWSXRayRecorderBenchmark {
         public void setupOnce() throws SocketException {
             recorder = AWSXRayRecorderBuilder.defaultRecorder();
         }
-
-        @Setup(Level.Invocation)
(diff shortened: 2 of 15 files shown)


aws-xray-recorder-sdk-benchmark/README.md
```diff
@@ -6,9 +6,10 @@ To run the benchmark, make sure you have Maven and Java 8 or above installed. Wi
 
 To run the benchmark, use your favorite shell to go into the directory. Then run the following commands:
 ```BASH
-cd aws-xray-sdk-java/aws-xray-recorder-sdk-benchmark/
-mvn clean && mvn package
-java -jar target/benchmark.jar
+git clone https://github.com/aws/aws-xray-sdk-java.git
+cd aws-xray-sdk-java
+./gradlew jmhJar
```
anuraaga, line 11: How about just `./gradlew jmh`?

aws-xray-recorder-sdk-benchmark/README.md
```diff
@@ -39,9 +40,9 @@ You should then start seeing the output of the benchmark:
 
 If you wish to run a specific benchmark, please run the following command:
 ```
-java -jar target/benchmarks.jar <Benchmark_name_here>
+java -jar aws-xray-recorder-sdk-benchmark-<VERSION>-jmh.jar <Benchmark_name_here>
```
anuraaga, line 43: Though it means forwarding a property into the Gradle config like here

https://github.com/line/armeria/blob/master/benchmarks/build.gradle#L32

aws-xray-recorder-sdk-benchmark/tst/main/java/com/amazonaws/xray/AWSXRayRecorderBenchmark.java
```diff
@@ -36,10 +36,10 @@
 import org.openjdk.jmh.runner.options.Options;
 import org.openjdk.jmh.runner.options.OptionsBuilder;
 
-@BenchmarkMode(Mode.All)
+@BenchmarkMode({Mode.Throughput, Mode.SampleTime})
```
anuraaga, line 39: Thanks, was thinking about this too!

aws-xray-recorder-sdk-benchmark/tst/main/java/com/amazonaws/xray/entities/EntityBenchmark.java
```diff
@@ -69,57 +78,42 @@ public void doTearDown() {
 
     // Construct a segment
     @Benchmark
-    @BenchmarkMode(Mode.All)
-    @Fork(value=1)
-    @Warmup(iterations = 20)
-    @Measurement(iterations = 20)
-    @OutputTimeUnit(TimeUnit.NANOSECONDS)
     public Segment constructSegmentBenchmark(BenchmarkState state) {
```
anuraaga, line 81: While we're cleaning, can we look for benchmarks using `Level.Invocation` and remove them? They're not real benchmarks for us since it's designed for much longer-running invocations (guessing when the benchmarks were added someone didn't know about the dragons).

http://javadox.com/org.openjdk.jmh/jmh-core/1.7/org/openjdk/jmh/annotations/Level.html#Invocation

build.gradle.kts
```diff
@@ -215,6 +215,11 @@ allprojects {
             dependsOn(publish)
         }
 
+        // Disable publishing a bunch of unnecessary Gradle metadata files
```
anuraaga, line 218:
```suggestion
        // Don't publish Gradle metadata for now until verifying they work well.
```

Technically they're not unnecessary for Gradle :) But it's true that since we need the POMs to work well anyways, they don't provide much value at least for now.

anuraaga (approved):

build.gradle.kts
willarmiros: Fair enough! I was just going off other AWS projects published to maven central and saw they didn't include the metadata. Will update comment.

aws-xray-recorder-sdk-benchmark/README.md
willarmiros: Yeah I tried this, and it gave lots of messy output where the gradle output of **[executing 1m 17s]** would interfere with the JMH output, and the JMH output would somehow spill over and appear when running `./gradlew build` afterwards.  I can include both options though in case it's just a me problem.

aws-xray-recorder-sdk-benchmark/tst/main/java/com/amazonaws/xray/entities/EntityBenchmark.java
willarmiros: I can definitely remove them, looks like `Level.Iteration` is more appropriate. Don't want to mess with any dragons.

aws-xray-recorder-sdk-benchmark/README.md
willarmiros: I think that's only if we were using the `jmh` task with gradle, when executing the jar directly I verified that just passing the fully qualified name of a benchmark class or test works.

aws-xray-recorder-sdk-benchmark/tst/main/java/com/amazonaws/xray/entities/EntityBenchmark.java
willarmiros: Actually, it turns out some of the tests need a per-invocation setup, e.g. to put a segment in context then end it, or clear segment context before beginning a new segment.

So we can either keep the necessary invocation rules, or scrap the benchmarks that require them since we already have full lifecycle benchmarks (e.g. only have `beginEndSegment` as opposed to having `beginSegment` and `endSegment` benchmarks). I'm in favor of the latter since some of the benchmarks can be quite redundant.

What do you think?

willarmiros: Discussed offline, will remove the non-lifecycle Recorder tests since it's very possible that they are not accurate measurements. Using a workaround with `OperationsPerInvocation` for entity benchmarks.

aws-xray-recorder-sdk-benchmark/tst/main/java/com/amazonaws/xray/entities/EntityBenchmark.java
```diff
@@ -51,75 +76,79 @@
...
-        @Setup(Level.Invocation)
-        public void setup() {
-            parentSegment = new SegmentImpl(recorder, SEGMENT_NAME);
+            for (int i = 0; i < N_OPERATIONS; i++) {
+                segments.add(new SegmentImpl(recorder, SEGMENT_NAME));
+            }
         }
 
         @TearDown(Level.Invocation)
         public void doTearDown() {
```
anuraaga, line 91: I think just running the setup on every iteration would be simpler than this, and avoids needing to add public API `setCause`

willarmiros: Running setup every iteration or invocation?

anuraaga: Sorry meant invocation (both start with i and end with n!)

willarmiros: Agreed!

PR #199, opened 2020-08-26
Updates changelog and benchmark for 2.7.0

willarmiros:
*Description of changes:*
I'll just leave this here for the pending minor version release.

Note that the benchmark measurements are only recorded for the two modes decided in #198 and no longer reflect non-lifecycle Recorder benchmarks, and we've added the trace header & some ID creation benchmarks.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,31 @@
 # Change Log
 
+## 2.7.0 - 2020-08-25
+- Add beginSegmentWithSampling API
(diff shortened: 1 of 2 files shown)


anuraaga (approved):
Thanks! I want to move this report into text files attached to the GitHub releases at some point to avoid an ever growing README.md. Let me know if you have any thoughts on that.

willarmiros:
Could not agree more, we did some work on automating running & recording benchmarks that didn't get finished, but we probably would've settled on an approach like that if we had finished.

Issue #216, opened 2020-09-24
Removing unnecessary SecureRandom use to improve ID generation performance


rhernandez35:
Recent profiling of my application shows I'm spending a significant amount of time generating random numbers for trace IDs. The RNG currently used is a cryptographically secure RNG, which is overkill for just generating unique identifiers. I'd like to improve performance by changing this to a plain ThreadLocalRandom.

PR #217, opened 2020-09-24
Improve ID generation performance

rhernandez35:
*Issue #, if available:*
https://github.com/aws/aws-xray-sdk-java/issues/216

*Description of changes:*
All segment and trace IDs are randomly generated with a SecureRandom when a normal Random would suffice. IDs do not need to be cryptographically secure, so the additional expense of generating a high-quality random sequence is wasted here. The simpler ThreadLocalRandom would provide a significant performance boost. In my test service, the X-Ray SDK was spending about 25% of its time here.


willarmiros:
Hi @rhernandez35 thanks for your contribution! We will discuss offline if this is the direction we'd like to go.

Issue #29, opened 2018-09-17
CentralizedSampling will not work with instrumentor


praetp:
When using XRay SDK 2.0.1, we see every 5 seconds: 
```
java.lang.NullPointerException: null
	at com.amazonaws.xray.handlers.TracingHandler.beforeExecution(TracingHandler.java:138) ~[aws-xray-recorder-sdk-aws-sdk-2.0.1.jar:?]
	at com.amazonaws.AmazonWebServiceClient.beforeClientExecution(AmazonWebServiceClient.java:723) ~[aws-java-sdk-core-1.11.398.jar:?]
	at com.amazonaws.services.xray.AWSXRayClient.getSamplingRules(AWSXRayClient.java:538) ~[aws-java-sdk-xray-1.11.398.jar:?]
	at com.amazonaws.xray.strategy.sampling.pollers.RulePoller.pollRule(RulePoller.java:53) ~[aws-xray-recorder-sdk-core-2.0.1.jar:?]
	at com.amazonaws.xray.strategy.sampling.pollers.RulePoller.lambda$start$0(RulePoller.java:41) ~[aws-xray-recorder-sdk-core-2.0.1.jar:?]
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511) [?:1.8.0_152]
	at java.util.concurrent.FutureTask.runAndReset(FutureTask.java:308) [?:1.8.0_152]
	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.access$301(ScheduledThreadPoolExecutor.java:180) [?:1.8.0_152]
	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:294) [?:1.8.0_152]
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149) [?:1.8.0_152]
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624) [?:1.8.0_152]
```

This is caused by the TracingHandler being created when the globalrecorder has not yet finished initialization. Then the recorder of the TracingHandler will be null. See stacktrace below. The globalrecorder is in the process of being created but AWSXray.globalRecorder has not been assigned yet.

When just getting the globalRecorder, you will trigger this problem:
```
[...]

PR #218, opened 2020-09-25
Add a configurable randomness source

rhernandez35:
I had two goals for this change:

1. Expose the configuration option without exposing any new classes or
interfaces. We should be in strict control of possible randomness
sources so we can guarantee that they actually do provide random values
and that they provide randomness efficiently. I didn't want it to be
possible for users to tank performance by inadvertently using a single
`Random` instance.
2. Make the random source configurable and scoped to each
`AWSXRayRecorder` instance. I didn't want to introduce any more global
shared state, so this unfortunately meant I wasn't able to remove the
global SecureRandom usage in two places: `TraceId#fromString` and
`DefaultThrowableSerializationStrategy#describeInContext`. The latter is
at least aware of the local segment context and will try to use the
random generator from its creator when possible, but the former is
unresolvable without either adding more global shared state or exposing
the `RandomGenerator` class.

This commit adds the following public APIs:

* `AWSXRayRecorder#useFastRandomGenerator` and
`AWSXRayRecorder#useSecureRandomGenerator`: for configuring the behavior
of each `AWSXRayRecorder` instance.
* `AWSXRayRecorderBuilder#useFastRandomGenerator` and
`AWSXRayRecorderBuilder#useSecureRandomGenerator`: same as above
* `ThrowableSerializationStrategy#describeInContext(SegmentContext,
Throwable, List<Subsegment>)`: a new method that allows the current
segment context to be provided to `ThrowableSerializationStrategy`. This
is currently only useful for getting the ID generator from the current
segment's creator, but I can imagine there being other uses for this
later.

I unfortunately had to add two internal-only public methods to
`AWSXRayRecorder` for actually getting random values from its configured
random generator. Adding methods there allowed me to keep the
`RandomGenerator` class hidden from users.

There are some other minor changes in this PR:

* Removed quadratic String concatenation in `Entity#generateId`
* `SegmentContextResolverChain` no longer double-resolves contexts

*Issue #, if available:*
#216 

### DIFF
--- a/aws-xray-recorder-sdk-core/src/main/java/com/amazonaws/xray/AWSXRayRecorder.java
+++ b/aws-xray-recorder-sdk-core/src/main/java/com/amazonaws/xray/AWSXRayRecorder.java
@@ -29,6 +29,9 @@ import com.amazonaws.xray.entities.Subsegment;
 import com.amazonaws.xray.entities.TraceID;
 import com.amazonaws.xray.exceptions.SegmentNotFoundException;
 import com.amazonaws.xray.exceptions.SubsegmentNotFoundException;
+import com.amazonaws.xray.internal.FastIdGenerator;
+import com.amazonaws.xray.internal.IdGenerator;
+import com.amazonaws.xray.internal.SecureIdGenerator;
 import com.amazonaws.xray.listeners.SegmentListener;
 import com.amazonaws.xray.strategy.ContextMissingStrategy;
 import com.amazonaws.xray.strategy.DefaultContextMissingStrategy;
@@ -117,6 +120,7 @@ public class AWSXRayRecorder {
     private PrioritizationStrategy prioritizationStrategy;
     private ThrowableSerializationStrategy throwableSerializationStrategy;
     private ContextMissingStrategy contextMissingStrategy;
+    private IdGenerator idGenerator;
 
     private SegmentContextResolverChain segmentContextResolverChain;
 
@@ -137,6 +141,7 @@ public class AWSXRayRecorder {
         prioritizationStrategy = new DefaultPrioritizationStrategy();
         throwableSerializationStrategy = new DefaultThrowableSerializationStrategy();
         contextMissingStrategy = new DefaultContextMissingStrategy();
+        idGenerator = new SecureIdGenerator();
 
         logReferences = new HashSet<>();
 
@@ -940,6 +945,39 @@ public class AWSXRayRecorder {
      */
     public void setOrigin(String origin) {
         this.origin = origin;
+    }
+
+    /**
+     * Configures this {@code AWSXRayRecorder} to use a fast but cryptographically insecure random number
+     * generator for generating random IDs. This option should be preferred if your application does not
+     * rely on AWS X-Ray Trace IDs being generated from a cryptographically secure random number generator.
+     *
+     * @see #useSecureIdGenerator()
+     */
+    public final void useFastIdGenerator() {
+        this.idGenerator = new FastIdGenerator();
+    }
+
+    /**
+     * Configures this {@code AWSXRayRecorder} to use a cryptographically secure random generator for
+     * generating random IDs. Unless your application in some way relies on AWS X-Ray trace IDs
+     * being generated from a cryptographically secure random number source, you should prefer
+     * to use {@linkplain #useFastIdGenerator() the fast ID generator}.
+     *
+     * @see #useFastIdGenerator()
+     */
+    public final void useSecureIdGenerator() {
+        this.idGenerator = new SecureIdGenerator();
+    }
+
+    /**
+     * Gets this {@code AWSXRayRecorder} instance's ID generator. This method is intended for
+     * internal use only.
+     *
+     * @return the configured ID generator
+     */
+    public final IdGenerator getIdGenerator() {
+        return idGenerator;
     }
 
     /**
--- a/aws-xray-recorder-sdk-core/src/main/java/com/amazonaws/xray/AWSXRayRecorderBuilder.java
+++ b/aws-xray-recorder-sdk-core/src/main/java/com/amazonaws/xray/AWSXRayRecorderBuilder.java
@@ -80,6 +80,8 @@ public class AWSXRayRecorderBuilder {
     @Nullable
     private Emitter emitter;
 
+    private boolean useFastIdGenerator = false;
+
 
     private AWSXRayRecorderBuilder() {
         plugins = new HashSet<>();
@@ -221,6 +223,31 @@ public class AWSXRayRecorderBuilder {
     }
 
     /**
+     * Prepares this builder to build an {@code AWSXRayRecorder} which uses a fast but cryptographically insecure
+     * random number generator for generating random IDs. This option should be preferred unless your application
+     * relies on AWS X-Ray trace IDs being generated from a cryptographically secure random number source.
+     *
+     * @see #withSecureIdGenerator() ()
+     */
+    public AWSXRayRecorderBuilder withFastIdGenerator() {
+        this.useFastIdGenerator = true;
+        return this;
+    }
+
+    /**
+     * Prepares this builder to build an {@code AWSXRayRecorder} which uses a cryptographically secure random
+     * generator for generating random IDs. Unless your application relies on AWS X-Ray trace IDs
+     * being generated from a cryptographically secure random number source, you should prefer
+     * to use {@linkplain #withFastIdGenerator() the fast ID generator}.
+     *
+     * @see #withFastIdGenerator()
+     */
+    public AWSXRayRecorderBuilder withSecureIdGenerator() {
+        this.useFastIdGenerator = false;
+        return this;
+    }
+
+    /**
      * Constructs and returns an AWSXRayRecorder with the provided configuration.
      *
      * @return a configured instance of AWSXRayRecorder
@@ -256,6 +283,12 @@ public class AWSXRayRecorderBuilder {
 
         if (!segmentListeners.isEmpty()) {
             client.addAllSegmentListeners(segmentListeners);
+        }
+
+        if (useFastIdGenerator) {
+            client.useFastIdGenerator();
+        } else {
+            client.useSecureIdGenerator();
         }
 
         plugins.stream().filter(Objects::nonNull).filter(p -> p.isEnabled()).forEach(plugin -> {
--- a/aws-xray-recorder-sdk-core/src/main/java/com/amazonaws/xray/ThreadLocalStorage.java
+++ b/aws-xray-recorder-sdk-core/src/main/java/com/amazonaws/xray/ThreadLocalStorage.java
@@ -59,6 +59,7 @@ public class ThreadLocalStorage {
         CURRENT_ENTITY.remove();
     }
 
+    @Deprecated
     public static SecureRandom getRandom() {
         return SECURE_RANDOM;
     }
--- a/aws-xray-recorder-sdk-core/src/main/java/com/amazonaws/xray/strategy/DefaultThrowableSerializationStrategy.java
+++ b/aws-xray-recorder-sdk-core/src/main/java/com/amazonaws/xray/strategy/DefaultThrowableSerializationStrategy.java
@@ -16,15 +16,18 @@
 package com.amazonaws.xray.strategy;
 
 import com.amazonaws.AmazonServiceException;
+import com.amazonaws.xray.AWSXRay;
 import com.amazonaws.xray.entities.Entity;
 import com.amazonaws.xray.entities.Subsegment;
 import com.amazonaws.xray.entities.ThrowableDescription;
+import com.amazonaws.xray.internal.IdGenerator;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 import java.util.Optional;
 import java.util.Set;
+import org.checkerframework.checker.nullness.qual.Nullable;
 
 /**
  * Default implementation of {@code ThrowableSerializationStrategy}.
@@ -109,7 +112,27 @@ public class DefaultThrowableSerializationStrategy implements ThrowableSerializa
 
     @Override
     public List<ThrowableDescription> describeInContext(Throwable throwable, List<Subsegment> subsegments) {
+        return describeInContext(null, throwable, subsegments);
+    }
+
+    @Override
+    public List<ThrowableDescription> describeInContext(
+        @Nullable Entity entity,
+        Throwable throwable,
+        List<Subsegment> subsegments
+    ) {
         List<ThrowableDescription> result = new ArrayList<>();
+        IdGenerator idGenerator;
+
+        // We could fall back on ThreadLocalStorage#get if we were confident that no `SegmentContext` implementations
+        // override the default get/setTraceEntity implementations, but we can't make that guarantee.
+        if (entity != null) {
+            idGenerator = entity.getCreator().getIdGenerator();
+        } else if (!subsegments.isEmpty()) {
+            idGenerator = subsegments.get(0).getCreator().getIdGenerator();
+        } else {
+            idGenerator = AWSXRay.getGlobalRecorder().getIdGenerator();
+        }
 
         /*
          * Visit each node in the cause chain. For each node:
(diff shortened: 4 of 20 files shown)


aws-xray-recorder-sdk-core/src/main/java/com/amazonaws/xray/strategy/ThrowableSerializationStrategy.java
```diff
@@ -32,4 +34,25 @@
...
+     * @param context
+     *            the {@link SegmentContext} for the current segment. May be null.
+     * @param throwable
+     *            the Throwable to serialize
+     * @param subsegments
+     *            the list of subsegment children in which to look for the same {@code Throwable} object, for chaining
+     *
+     * @return a list of {@code ThrowableDescription}s which represent the provided {@code Throwable}
+     */
+    default List<ThrowableDescription> describeInContext(
```
willarmiros, line 51: Just to play devil's advocate since this is a fairly tedious change and new API, do we need the performance bump in the throwable serializer? It's only used for requests that are already "bad" by nature of having an exception that needs recording. 

I'm ok with leaving it, especially if you think the performance is still important here, but will leave it up to you.

willarmiros: Actually, looking at it again, is the segment context (and this new API) necessary at all? Can't you just grab a subsegment out of the list in the original API, do `getCreator()` on it and use the `generateId` method there?

aws-xray-recorder-sdk-core/src/main/java/com/amazonaws/xray/AWSXRayRecorder.java
```diff
@@ -1029,4 +1054,24 @@ public String currentFormattedId() {
             return null;
         }
     }
+
+    /**
+     * Internal-only utility method.
+     */
+    @Deprecated
+    public final <T> T getRandomBytes(int bytes, Function<byte[], T> mapper) {
```
willarmiros, line 1062: I know @rhernandez35 you want to avoid exposing these, but I'm not sure how I feel about adding new internal-only methods to an already very bloated Recorder class. IMO these methods can be extracted into a public interface/class (like `RandomGenerator`) and exposed to customers. That is to some degree what we're requesting OpenTelemetry, a pluggable ID generation experience. I'm not sure if customers would bother using it unless they really knew what they were doing, and we can still guarantee randomness with our *own* generators.

@anuraaga I'd appreciate you weighing in on this too, I could be completely off the mark here.

willarmiros (commented):
Thanks again for the changes! Left some thoughts and we should prob wait to hear from @anuraaga before getting this in.

aws-xray-recorder-sdk-core/src/main/java/com/amazonaws/xray/AWSXRayRecorder.java
rhernandez35: If you're fine with increasing the surface area then I'll gladly do it since the code would be so much cleaner. However, my goal was to avoid providing customers a way to evade our curated selection of randomness providers. I really don't want someone to think "oh hey I can save memory by just using `Math#random`" and accidentally add a global synchronization point on every new ID creation.

Adding these methods here was definitely a gigantic hack, but it seemed like the lesser of two evils.

aws-xray-recorder-sdk-core/src/main/java/com/amazonaws/xray/strategy/ThrowableSerializationStrategy.java
rhernandez35: I did that at first, but there's no guarantee that the list will contain any subsegments. In my simple demo service, we just have the one master segment. I could search the list as a fallback but I didn't think that was necessary since I'm hoping most users are using the default serializer, which is able to always provide the segment in use.

I also experimented with adding an `Entity` parameter. Maybe I should have kept that? It was definitely safer – the only way to reach this method in X-Ray's codebase is through `EntityImpl#addException`, and while extremely unlikely, there's no guarantee that the entity invoking `describeInContext` is the same entity that's stored by the segment context. This wasn't a problem with the `Entity` parameter because `EntityImpl` could just give `this` as an argument.

aws-xray-recorder-sdk-core/src/main/java/com/amazonaws/xray/AWSXRayRecorderBuilder.java
```diff
@@ -220,6 +222,31 @@ public AWSXRayRecorderBuilder withDefaultPlugins() {
...
     }
 
+    /**
+     * Prepares this builder to build an {@code AWSXRayRecorder} which uses a fast but cryptographically insecure
+     * random number generator for generating random IDs. This option should always be used unless your application in
+     * some way relies on AWS X-Ray trace IDs being generated from a cryptographically secure random number source.
+     *
+     * @see #useSecureRandomGenerator()
+     */
+    public AWSXRayRecorderBuilder useFastRandomGenerator() {
```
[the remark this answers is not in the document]
rhernandez35: indeed we can

aws-xray-recorder-sdk-core/src/main/java/com/amazonaws/xray/AWSXRayRecorder.java
anuraaga: I lean towards having only two types of random - OpenTelemetry supports trace ID customization since it needs to support all backends, including weird ones like X-Ray. X-Ray SDK on the flip side can't be used with any other backend so we can just limit the options.

Instead of the pattern of deprecated method for internal API, we have the `.internal` package - can we move he internal pieces there?

anuraaga (commented):
Sorry for the late review - left a high level comment about the internal API. Thanks a lot for the help @rhernandez35!

aws-xray-recorder-sdk-core/src/main/java/com/amazonaws/xray/AWSXRayRecorder.java
rhernandez35: sure, I'm fine with that if you are.

I can say from experience though that users don't really care what the fully qualified name for a class is when they use it. If it's pubic and their IDE can autoresolve it, they often don't even know what package it's in. If it's public, it _will_ be used in ways you can't anticipate.

willarmiros: I think the `internal` package sounds good to me. Sure customers *can* override it if they want, but it will not be a publicly documented configuration like the SDK's other configs (as a matter of fact the JavaDoc should specifically call out that it's an internal-only API that is not meant to be used by consumers of the SDK, which may help guide some of those IDE users 😉 ).

rhernandez35:
OK, I moved the random generators to `internal` and added a getter in `AWSXRayRecorder`. Everything else is mostly the same as in the last revision.

aws-xray-recorder-sdk-core/src/main/java/com/amazonaws/xray/utils/SecureRandomSupplier.java
```diff
@@ -0,0 +1,63 @@
...
+/**
+ * An <em>internal</em> utility class to efficiently generate {@code SecureRandom} instances. This class
+ * should not be used outside of the AWS X-Ray SDK.
+ */
+public final class SecureRandomSupplier implements Supplier<SecureRandom> {
+    private static final Supplier<SecureRandom> RANDOM_SUPPLIER;
+    public static final SecureRandomSupplier INSTANCE = new SecureRandomSupplier();
+
+    static {
+        // JDK 9+: Use a thread-local SecureRandom instance if we're able to determine
```
anuraaga, line 31: We removed threadlocal securerandom here after adding benchmarks for it in #162

I believe it's because the OS's random is guaranteed to be a global lock, so a thread-local securerandom instance doesn't reduce locking but does add overhead for the threadlocal lookup.

If this makes sense (maybe my benchmark is missing something, never know), could you revert this change?

aws-xray-recorder-sdk-core/src/main/java/com/amazonaws/xray/internal/RandomGenerator.java
```diff
@@ -0,0 +1,44 @@
...
+ * depend on or use this class.
+ */
+public abstract class RandomGenerator {
+    private final Supplier<Random> randomSupplier;
+
+    protected RandomGenerator(Supplier<Random> randomSupplier) {
+        this.randomSupplier = randomSupplier;
+    }
+
+    public final byte[] randomBytes(int bytes) {
```
anuraaga, line 35: Since this is an internal class we can tie it to our requirements - let's have it return the base16 string itself. For secure random, `nextBytes` is much faster but for normal random, two `nextLong`s is faster, we should take advantage of that. If it's too much for this PR, we can address it later though no worries.

aws-xray-recorder-sdk-core/src/main/java/com/amazonaws/xray/AWSXRayRecorder.java
```diff
@@ -942,6 +947,39 @@ public void setOrigin(String origin) {
         this.origin = origin;
     }
 
+    /**
+     * Configures this {@code AWSXRayRecorder} to use a fast but cryptographically insecure random number
+     * generator for generating random IDs. This option should always be used unless your application in
```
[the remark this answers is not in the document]
rhernandez35: Ok, I'll tone it down a bit.

aws-xray-recorder-sdk-core/src/main/java/com/amazonaws/xray/utils/SecureRandomSupplier.java
rhernandez35: the entropy pool is just one source of blocking and actually isn't directly queried all that often (usually only for seed generation). the main source of contention I'm worried about is the thread-safety of the RNG algorithm itself – RNGs typically maintain internal state that prevent concurrent reads/updates. This blob of code checks for when that's true and uses thread locals as necessary.

what os/arch did you run your benchmarks on? what crypto provider(s) did you install? which JVM – corretto? [AmazonCorrettoCryptoProvider](https://aws.amazon.com/about-aws/whats-new/2019/07/introducing-the-amazon-corretto-crypto-provider/) provides a thread-safe `SecureRandom` implementation that stores its state in thread-locals, so `SecureRandom` skips synchronizing on calls into it. You'll see lock contention in SecureRandom if you use a crypto provider that doesn't provide a thread-safe random.

rhernandez35: now that I think about it, it would probably be better to always use a thread-local SecureRandom since "thread-safe" could also be implemented by just slapping `synchronized` on your random spi's methods.

anuraaga: It sounds like coretto would be even faster without the threadlocal. But I just tested with normal openjdk 14 on Mac and even with that, not using threadlocal was a bit faster. I don't know how well my theory checks out, but since the benchmark numbers aren't convincing and threadlocal securerandom adds complexity, let's go ahead and stick with the singleton please. Thanks!

aws-xray-recorder-sdk-core/src/main/java/com/amazonaws/xray/strategy/ThrowableSerializationStrategy.java
willarmiros: Ok yeah I like passing the entity instead of the context in that case - if we're just using the context to retrieve the entity then it definitely sounds simpler and safer to pass in the entity.

willarmiros (commented):
Couple more comments

aws-xray-recorder-sdk-core/src/main/java/com/amazonaws/xray/ThreadLocalStorage.java
```diff
@@ -24,9 +25,7 @@
  */
 @SuppressWarnings("checkstyle:HideUtilityClassConstructor")
 @Deprecated
-public class ThreadLocalStorage {
-
-    private static final SecureRandom SECURE_RANDOM = new SecureRandom();
+public final class ThreadLocalStorage {
```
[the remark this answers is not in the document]
rhernandez35: I suppose, but this is documented as an internal-only class. I thought y'all were ok with making breaking changes to internal code?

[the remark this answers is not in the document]
anuraaga: Ah good catch - yeah let's not change to final this time. Internal was added relatively recently in the hopes of significant cleanup in a major version like hiding this class. So the internal package is ok, all classes there started out as that, but not anything outside. If this code generally looks like a mess, yes you're right :)

willarmiros (approved):
LGTM, will wait on @anuraaga to approve and merge. Thanks for your contribution and patience @rhernandez35!

aws-xray-recorder-sdk-core/src/main/java/com/amazonaws/xray/strategy/DefaultThrowableSerializationStrategy.java
```diff
@@ -109,7 +112,29 @@ private ThrowableDescription describeThrowable(Throwable throwable, String id) {
...
+        // We could fall back on ThreadLocalStorage#get if we were confident that no `SegmentContext` implementations
+        // override the default get/setTraceEntity implementations, but we can't make that guarantee.
+        if (entity != null) {
+            idGenerator = entity.getCreator().getIdGenerator();
+        } else if (!subsegments.isEmpty()) {
+            idGenerator = subsegments.get(0).getCreator().getIdGenerator();
+        } else {
+            // to maintain backwards compatibility, always fall back to the secure ID generator when we can't
+            // find a different one set
+            idGenerator = SecureIdGenerator.INSTANCE;
```
rhernandez35, line 136: how would you feel about me using the ID generator from the global recorder (via `AWSXRay::getGlobalRecorder`) instead of this?

rhernandez35: and same for in `TraceID::fromString`/`TraceID::create`

willarmiros: Ah I'd actually prefer that! Since that will also default to secure if the user hasn't configured their recorder, but at least has a chance of picking up the configured value if the first two conditions don't get hit somehow.

rhernandez35: Are users allowed to set the global recorder to `null`? Most code assumes it won't return null, but there's no explicit check in `setGlobalRecorder`. Further, there's a test that even sets it to null, alongside some code in `TracingHandler` that's meant to deal with `getGlobalRecorder` returning null (although this code also assumes that all subsequent invocations of `getGlobalRecorder` will return an object).

willarmiros: We added `Nullable` annotations throughout the SDK and left the `globalRecorder` field and corresponding getter in `AWSXRay` as not `Nullable` so I think it's safe to assume that customers are respecting the non-null contract. In `TracingHandler` I think we're protecting against the `TracingHandler` being constructed with a `null` recorder parameter, and then assume that `getGlobalRecorder` will provide the recorder if so. What test sets it to null?

rhernandez35: `testRaceConditionOnRecorderInitialization` in `TracingHandlerTest`

willarmiros: Ah yeah that's unrelated to testing if the customer has set the global recorder to null, it's [this issue](https://github.com/aws/aws-xray-sdk-java/issues/29) where the AWS SDK's auto-instrumentor would create a `TracingHandler` and call `AWSXRay.getGlobalRecorder` in parallel with the recorder's start up before it had finished initializing. Not something that we'd expect in regular customer code that's run after the recorder is initialized.

anuraaga (approved):
Thanks @rhernandez35!

PR #221, opened 2020-10-05
Add workflow for release build.

anuraaga:
Not well verified yet, will probably create a testing release tag to see how it goes.

### DIFF
--- a/.github/workflows/release-build.yml
+++ b/.github/workflows/release-build.yml
@@ -1,0 +1,50 @@
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        description: The version to tag the release with, e.g., 1.2.0, 1.2.1-alpha.1
+        required: true
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


.github/workflows/release-build.yml
```diff
@@ -0,0 +1,50 @@
...
+          GRGIT_PASS: ${{ secrets.GITHUB_TOKEN }}
+      - name: Create Release
+        id: create_release
+        uses: actions/create-release@v1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          tag_name: v${{ github.event.inputs.version }}
+          release_name: Release v${{ github.event.inputs.version }}
+          draft: true
```
willarmiros, line 49: jw, what does `draft` do?

willarmiros (approved):
Awesome!

.github/workflows/release-build.yml
anuraaga: It means the release is not published so users that aren't maintainersof the repo can't see it yet. Figured letting that final step being opt-in, along with writing the release notes, is ok.

PR #222, opened 2020-10-05
Removed some powermock

willarmiros:
*Description of changes:*
Removed unnecessary powermock from ECS plugin tests. I tried using the JUnit Pioneer `SetEnvironmentVariable` annotation but for some reason couldn't get it to actually set the environment variable in the tests. Specifically, I tried:

```java
@Test
@SetEnvironmentVariable(key = ECS_METADATA_KEY, value = GOOD_URI)
```

but the environment variable would still be `null`, so I just went with System Rules instead.

### DIFF
--- a/aws-xray-recorder-sdk-core/src/test/java/com/amazonaws/xray/plugins/ECSPluginTest.java
+++ b/aws-xray-recorder-sdk-core/src/test/java/com/amazonaws/xray/plugins/ECSPluginTest.java
@@ -15,39 +15,41 @@
 
 package com.amazonaws.xray.plugins;
 
-import org.junit.Assert;
-import org.junit.Test;
-import org.junit.runner.RunWith;
-import org.powermock.api.mockito.PowerMockito;
-import org.powermock.core.classloader.annotations.PrepareForTest;
-import org.powermock.modules.junit4.PowerMockRunner;
+import static org.assertj.core.api.Assertions.assertThat;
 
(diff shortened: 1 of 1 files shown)


aws-xray-recorder-sdk-core/src/test/java/com/amazonaws/xray/plugins/ECSPluginTest.java
```diff
@@ -15,39 +15,41 @@
 
 package com.amazonaws.xray.plugins;
 
-import org.junit.Assert;
+import static org.assertj.core.api.Assertions.assertThat;
+
+import org.junit.Rule;
 import org.junit.Test;
```
anuraaga, line 21: Thanks for the cleanup! Did you try pioneer after changing this import to jupiter? Otherwise, I also wouldn't expect BeforEach to work properly, currently we have both JUnit4 and 5 here I think.

willarmiros: Ah I see, yeah let me try converting the whole class to JUnit 5

willarmiros: Works now, thanks for the tip!

anuraaga (approved):
Sweet

PR #224, opened 2020-10-07
Avoid nanosecond-resolution time where not needed

rhernandez35:
`Instant#now` returns an Instant with nanonsecond-level resolution. The
ns-level timer isn't cheap, and is always discarded in X-Ray's code. We
can avoid much of this cost by swapping out `Instant#now` with
`System#currentTimeMillis`.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/aws-xray-recorder-sdk-core/src/main/java/com/amazonaws/xray/internal/TimeUtils.java
+++ b/aws-xray-recorder-sdk-core/src/main/java/com/amazonaws/xray/internal/TimeUtils.java
@@ -1,0 +1,31 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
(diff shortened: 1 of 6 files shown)


aws-xray-recorder-sdk-core/src/main/java/com/amazonaws/xray/internal/TimeUtils.java
```diff
@@ -0,0 +1,31 @@
...
+    /**
+     * @return the current epoch second
+     */
+    public static long currentEpochSecond() {
+        return MILLISECONDS.toSeconds(System.currentTimeMillis());
+    }
+
+    private TimeUtils() {
+        throw new AssertionError("no instances, please");
+    }
```
anuraaga, line 30:
```suggestion
    private TimeUtils() {}
```

Think we don't need to add the extra noise

anuraaga (approved):
Thanks!

anuraaga:
Thanks @rhernandez35

Issue #230, opened 2020-11-06
ForcedSampling is not properly supported


friscoMad:
There is no way to enable ForcedSampling in any of the provided Strategies, we can extend our own strategy to enable it but it is far from ideal.
beginSegmentWithSampling does not take ForcedSampling into account, it is only supported by the servlet filter.
Ideally, the servlet filter should also support forcing sampling by status code (so for example all 500 errors are sampled).

willarmiros:
Hi @friscoMad,

Thanks for raising the fact that ForcedSampling is not taken into account for `beginSegmentWithSampling`, I can address that bug. We can also add a new constructor parameter to enable ForcedSampling on our currently vended strategies. At this time we don't plan on releasing any official sampling strategies that use forced sampling directly, but you are always able to implement custom strategies for such use cases.

PR #232, opened 2020-11-09
improved ForcedSampling support

willarmiros:
*Issue #, if available:*
#230 

*Description of changes:*
Adds check for `ForcedSamplingSupport` to `beginSegmentWithSampling`. Also makes `ForcedSampling` a configurable property for the existing `Centralized` and `Localized` sampling strategies via new constructors. I did not add them to the `All` or `No` strategies since forced sampling doesn't make sense in either case, they should be all or nothing, respectively.  

I used constructor parameters to keep the new field `final`. One alternative would be to implement builder classes for the strategies (4 new builder classes seems overkill tho) or just one sampling strategy builder that works like:

```java
SamplingStrategy strategy = SamplingStrategyBuilder
    .withStrategy("Centralized")  // Or withCentralizedSampling() ?
    .withRuleLocation(someUrl)  // These are no-ops for All and None strategies
    .withForcedSampling()
    .build();
```

Thoughts @anuraaga ?

### DIFF
--- a/aws-xray-recorder-sdk-core/src/main/java/com/amazonaws/xray/AWSXRayRecorder.java
+++ b/aws-xray-recorder-sdk-core/src/main/java/com/amazonaws/xray/AWSXRayRecorder.java
@@ -416,6 +416,14 @@ public class AWSXRayRecorder {
             }
 
             return segment;
+        } else if (this.getSamplingStrategy().isForcedSamplingSupported()) {
+            Segment segment = beginSegment(name);
+            segment.setSampled(false);
+            if (samplingResponse.getRuleName().isPresent()) {
+                segment.setRuleName(samplingResponse.getRuleName().get());
+            }
+
+            return segment;
         }
 
         return beginNoOpSegment();
--- a/aws-xray-recorder-sdk-core/src/main/java/com/amazonaws/xray/AWSXRayRecorder.java
+++ b/aws-xray-recorder-sdk-core/src/main/java/com/amazonaws/xray/AWSXRayRecorder.java
@@ -989,7 +989,8 @@ public class AWSXRayRecorder {
     }
 
     /**
-     * Checks whether the current {@code SamplingStrategy} supports forced sampling.
+     * Checks whether the current {@code SamplingStrategy} supports forced sampling. Use with caution, since segments sampled in
+     * this manner will not count towards your sampling statistic counts.
      *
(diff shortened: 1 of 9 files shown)


aws-xray-recorder-sdk-core/src/main/java/com/amazonaws/xray/AWSXRayRecorder.java
```diff
@@ -415,6 +415,14 @@ public Segment beginSegmentWithSampling(String name) {
                 segment.setRuleName(samplingResponse.getRuleName().get());
             }
 
+            return segment;
+        } else if (this.getSamplingStrategy().isForcedSamplingSupported()) {
+            Segment segment = beginSegment(name);
+            segment.setSampled(false);
```
srprash, line 421: From what I understand of ForcedSampling is that it can be used to sample a segment which is un-sampled when created. Or is it only to keep the segment in context even though it is un-sampled?

aws-xray-recorder-sdk-core/src/main/java/com/amazonaws/xray/strategy/sampling/CentralizedSamplingStrategy.java
```diff
@@ -149,7 +155,6 @@ private synchronized void startPoller() {
     @Override
     public boolean isForcedSamplingSupported() {
         //TODO address this
```
srprash, line 157: This is addressed now right? :)

willarmiros: Fair enough haha

aws-xray-recorder-sdk-core/src/main/java/com/amazonaws/xray/AWSXRayRecorder.java
willarmiros: Forced sampling forces the creation of a regular instead of no-op segment when not sampled, which also places it in context. Then `forceSamplingOfCurrentSegment` can be used in conjunction to actually switch an unsampled segment to sampled.

anuraaga (approved):
I think the constructor parameters are OK - the builder seems nice if we were to also hide the classes, but that would be breaking, so maybe a future cleanup

PR #242, opened 2020-11-25
added aggregate javadoc plugin

willarmiros:
*Description of changes:*
We need to aggregate all javadocs at the top-level to publish them publicly. This tool generates them with:

```
./gradlew aggregateJavadoc
```

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -11,7 +11,7 @@ plugins {
 
     id("nebula.release")
     id("org.ajoberstar.grgit")
-    id("io.freefair.aggregate-javadoc")
+    id("io.freefair.aggregate-javadoc-jar")
 }
 
 tasks {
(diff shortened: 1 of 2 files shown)


srprash (approved):

willarmiros:
Switched to plugin for aggregate javadoc jar, which does the same javadoc generation as the previous plugin, but also allows for the ability to create an aggregate jar with `./gradlew aggregateJavadocJar`, which is what our system actually needs.

anuraaga (approved):
Just wondering, how do we publish the javadoc? Is it something we can add to the release workflow?

willarmiros:
@anuraaga I don't think so unfortunately, the javadocs are currently deployed to an internally-managed host. We can discuss today about their current architecture.

PR #243, opened 2020-11-30
Add option to  always create trace ID 

### LINKED ISSUE #238: Add configuration option to create traceIds for NoOpSegments
The default behavior when a SamplingResponse is NotSampled is to create a NoOpSegment with TraceID.invalid() as the traceId, in our use case we would like the traceId to be created even in that case so it could be propagated to other services.

Our use case is that we are using the traced as part of the context in our logging system and use that to filter all logs from a single request even when they are from different services. (We are currently using openzipkin / brave).

Currently the servletFilter supports reading than passing the traceId if it comes in the header from other services even with Sampled=0.
There is a `beginSegment(String name, TraceID traceId, @Nullable String parentId)` that could be used as a workaround if we wouldn't like to use sampling but there is nothing that could be used with beginSegmentWithSampling().

I propose a new recorder option and the following change:
```
    public Segment beginNoOpSegment() {
        return beginSegment(Segment.noOp(this.forceCreateTrace?TraceID.create():TraceID.invalid(), this));
    }
```
I think that's all it is needed to fully support this new feature.

friscoMad:
*Issue #, if available:*
Fixes #238 

*Description of changes:*
This PR adds a new method in the builder and in the recorder to force NoOpSegments to have a valid individual Trace ID.
I am not entirely happy with the variable name so suggestions are accepted.

### DIFF
--- a/aws-xray-recorder-sdk-core/src/main/java/com/amazonaws/xray/AWSXRayRecorder.java
+++ b/aws-xray-recorder-sdk-core/src/main/java/com/amazonaws/xray/AWSXRayRecorder.java
@@ -135,6 +135,7 @@ public class AWSXRayRecorder {
 
     @MonotonicNonNull
     private String origin;
+    private boolean alwaysCreateTraceId;
 
     public AWSXRayRecorder() {
         samplingStrategy = new DefaultSamplingStrategy();
@@ -441,7 +442,7 @@ public class AWSXRayRecorder {
      * will be propagated downstream.
      */
     public Segment beginNoOpSegment() {
-        return beginSegment(Segment.noOp(TraceID.invalid(), this));
+        return beginSegment(Segment.noOp(this.alwaysCreateTraceId ? TraceID.create(this) : TraceID.invalid(), this));
     }
 
     /**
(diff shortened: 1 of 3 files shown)


aws-xray-recorder-sdk-core/src/main/java/com/amazonaws/xray/AWSXRayRecorderBuilder.java
```diff
@@ -247,6 +248,17 @@ public AWSXRayRecorderBuilder withSecureIdGenerator() {
         return this;
     }
 
+    /**
+     * Prepares this builder to build an {@code AWSXRayRecorder} which creates Trace ID for all Segments
+     * even for NoOpSegments or not sampled ones that usually include include a static invalid TraceID.
+     * This could be useful for example in case the Trace ID is logged to be able to aggregate all logs from a
+     * single request
+     */
+    public AWSXRayRecorderBuilder withAlwaysCreateTraceId() {
```
willarmiros, line 257: I think `withForcedTraceIdGeneration` would sound better.

aws-xray-recorder-sdk-core/src/main/java/com/amazonaws/xray/AWSXRayRecorder.java
```diff
@@ -1082,4 +1083,12 @@ public String currentFormattedId() {
             return null;
         }
     }
+
+    /**
+     * Configures this {@code AWSXRayRecorder} to add valid TraceId in all segments even NoOp ones that usually have
+     * a fixed value.
+     */
+    public void alwaysCreateTraceID(final boolean alwaysCreateTraceId) {
```
willarmiros, line 1091: It would probably be best if this method was explicitly named as a setter, e.g. `setForcedTraceIdGeneration`.

willarmiros (commented):
Thanks for this contribution! LGTM overall, just a naming change recommendation as requested.

willarmiros (approved):
Thanks!

Issue #239, opened 2020-11-19
Centralized Sampling rules are not removed when they are deleted


friscoMad:
I found this during my test, deleted a rule in the control panel but it was not refreshed in the app even after several "Polling sampling rules."
I did a bit of investigation and I think that when the RulePoller does retrieve the list of rules they are passed to the CentralizedManifest and there, only changes in priority or new rules trigger a full invalidation, otherwise, rules are updated but never deleted (even when the comment seems to indicate the contrary).

lupengamzn:
Hey @friscoMad ,

Thanks for pointing this out and contributing, seems like this is indeed an issue and we'll review your PR soon.

PR #244, opened 2020-11-30
Fix Sampling rules not being deleted

friscoMad:
*Issue #, if available:*
#239 

*Description of changes:*
Review if any of the rules present in the manifest is not present in the payload to rebuild the manifest (I am not sure if removing just that rule is enough or we need to rebuild so I opted for rebuilding). 

### DIFF
--- a/aws-xray-recorder-sdk-core/src/main/java/com/amazonaws/xray/strategy/sampling/manifest/CentralizedManifest.java
+++ b/aws-xray-recorder-sdk-core/src/main/java/com/amazonaws/xray/strategy/sampling/manifest/CentralizedManifest.java
@@ -93,12 +93,23 @@ public class CentralizedManifest implements Manifest {
         boolean invalidate = false;
 
         Map<String, CentralizedRule> rules = this.rules;
+        List<String> inputNames = new ArrayList<>(inputs.size());
 
         for (SamplingRule i : inputs) {
             if (i.getRuleName().equals(CentralizedRule.DEFAULT_RULE_NAME)) {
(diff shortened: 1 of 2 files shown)


aws-xray-recorder-sdk-core/src/main/java/com/amazonaws/xray/strategy/sampling/manifest/CentralizedManifest.java
```diff
@@ -93,14 +93,25 @@ public void putRules(List<SamplingRule> inputs, Instant now) {
...
         for (SamplingRule i : inputs) {
             if (i.getRuleName().equals(CentralizedRule.DEFAULT_RULE_NAME)) {
                 putDefaultRule(i);
             } else {
+                inputNames.add(i.getRuleName());
                 invalidate = putCustomRule(rules, i);
             }
         }
+        // Check if any rule was removed
+        if (!invalidate) {
```
wangzlei, line 107: What would happen if inputs has more than one items but the last item gets true?

friscoMad: I don't understand your question. 
If we receive a set with a new rule invalidate will be true and the whole set will be rebuilt so we don't need to do anything, otherwise, we check if there is a rule that is no longer present and if that's the case these changes trigger a rebuild to remove it and reconstruct the rule priorities.

wangzlei (approved):

Issue #108, opened 2019-11-07
Make segment/subsegment closing thread-safe 


shengxil:
When closing a subsegment, we check if its parent segment is completed and has no other subsegment. If so we will emit the segment as well. This is because when we close a segment which parents live subsegments, we don't emit it but wait for all child subsegments complete. However, the current implementation is not thread-safe per below:

Customer ask:
"
Looking at the AWS X-Ray Java SDK code:

* https://github.com/aws/aws-xray-sdk-java/blob/master/aws-xray-recorder-sdk-core/src/main/java/com/amazonaws/xray/entities/EntityImpl.java#L496
* https://github.com/aws/aws-xray-sdk-java/blob/master/aws-xray-recorder-sdk-core/src/main/java/com/amazonaws/xray/entities/SegmentImpl.java#L53
The problems here are that (a) `LongAdder` is something of an "eventually consistent" counter and (b) checking `isInProgress` and `referenceCount` doesn't happen atomically. The end result is that both the segment and the subsegment itself may attempt to emit the subsegment.
To fix this, the following changes are needed:

* `referenceCount` must be turned into an `AtomicInteger` instead of `LongAdder`.
* The reference count must include the segment itself, i.e. it must start at 1 and reach zero when the segment has been closed. With this change, the separate `!isInProgress()` check can and must be removed.
* `Decrementing` and checking the value must be done atomically, i.e. `AtomicInteger#decrementAndGet()` instead of `LongAdder#decrement() followed by LongAdded#intValue()`.
"

anoopgarlapati:
Hi @shengxil 
I'm facing a problem in my application for some of the traces for multi-threaded requests and I'm trying to understand if this is the same problem. Please excuse me if it is unrelated to this issue, I will raise it separately.

The problem is I'm receiving the following error log for some of the traces for multi-threaded requests.
```
2019-11-27 08:46:27.416 ERROR 3229 --- [ool-6-thread-24] c.a.x.e.EntityImpl                       : Exception while serializing entity.

com.fasterxml.jackson.databind.JsonMappingException: Index: 76, Size: 75 (through reference chain: com.amazonaws.xray.entities.SegmentImpl["subsegments"]->java.util.ArrayList[76])
	at com.fasterxml.jackson.databind.JsonMappingException.wrapWithPath(JsonMappingException.java:394) ~[jackson-databind-2.10.1.jar!/:2.10.1]
[...]

PR #250, opened 2021-01-19
Lock all accesses to entities.

anuraaga:
*Issue #, if available:* #108

*Description of changes:*
Almost none of the entity accesses are currently safe since there are many mutations of non-volatile fields. While making fields volatile is one way to help, modern CPUs have fast enough performance for non-contended locks (just atomic CAS) that it's common to just use locks (Golang makes it very hard to use volatile for this reason). This is the implementation OTel uses

https://github.com/open-telemetry/opentelemetry-java/blob/master/sdk/trace/src/main/java/io/opentelemetry/sdk/trace/RecordEventsReadableSpan.java

OTel doesn't have setters for every single thing so doesn't need as much locking, but as we don't expect much contention here it should be fine. 

I haven't quite reasoned through whether this fixes the subsegment issues, but it's an important change regardless.

### DIFF
--- a/aws-xray-recorder-sdk-core/src/main/java/com/amazonaws/xray/entities/EntityImpl.java
+++ b/aws-xray-recorder-sdk-core/src/main/java/com/amazonaws/xray/entities/EntityImpl.java
@@ -33,10 +33,11 @@ import com.fasterxml.jackson.databind.module.SimpleModule;
 import com.fasterxml.jackson.databind.node.NullNode;
 import com.fasterxml.jackson.databind.ser.BeanSerializerModifier;
 import com.fasterxml.jackson.databind.ser.std.ToStringSerializer;
+import com.google.errorprone.annotations.concurrent.GuardedBy;
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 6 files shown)


aws-xray-recorder-sdk-core/src/main/java/com/amazonaws/xray/entities/EntityImpl.java
```diff
@@ -586,39 +722,47 @@ public boolean isEmitted() {
...
 
     @Override
     public void removeSubsegment(Subsegment subsegment) {
-        getSubsegmentsLock().lock();
-        try {
-            subsegments.remove(subsegment);
-        } finally {
-            getSubsegmentsLock().unlock();
+        synchronized (lock) {
+            getSubsegmentsLock().lock();
```
willarmiros, line 758: Do we still need to lock the subsegments list when we're locking on the entire parent entity?

willarmiros (approved):
LGTM, we'll need to keep an eye on benchmarks before next release but shouldn't be a big deal. My guess is this should resolve #108 since the issue there was that during serialization, a subsegment could be removed from the entity being serialized. Since the serialize method and the methods to add/remove subsegments are controlled by the same lock now, it should be ok now. But can't be sure without reliable repro case/test.

PR #261, opened 2021-02-21
Generate and upload code coverage report on PRs

srprash:
Using the Jacoco plugin for gradle, generating an aggregated code coverage report. The report will be generated in the `jacoco` project and the `pr-build` workflow will upload the report to codecov.io.
Since we need only one report to be generated, I'm doing so in the build for Java 11.

For this to work completely, we need two more things:
1. The administrator needs to login/sign-up on codecov.io and add the this repo to the list of repositories so that the codecov/codecov-action can upload the coverage report
2. Add the codecov bot to the repo so that it can post the coverage diff as a comment on pull requests.

### DIFF
--- a/.github/workflows/master-build.yml
+++ b/.github/workflows/master-build.yml
@@ -17,28 +17,48 @@ jobs:
         java:
           - 8
           - 11
+        include:
+          - os: ubuntu-latest
+            java: 11
+            coverage: true
     steps:
       - name: Checkout Repository
         uses: actions/checkout@v1
+
       - name: Setup java
         uses: actions/setup-java@v1
         with:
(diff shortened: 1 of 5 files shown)


anuraaga (commented):
I think for OSS repos it should basically just work without extra steps but let's see.

Can you add to main build too? codecov.io needs the main branch uploaded too to calculate coverage diffs

srprash:
Yeah, it seems to be working without those steps. I had to add the codecov to my fork so I thought it would be required here as well.

anuraaga (approved):

willarmiros (approved):
Awesome addition!

PR #267, opened 2021-03-22
Added JavaTimeModule for entity object mapper

willarmiros:
*Description of changes:*
Adds a `JavaTimeModule` to our ObjectMapper for serializing entities, so that we can support `Duration` and other Time-based attributes being added to Metadata. We had a customer feature request for this

@anuraaga I know the tests will fail for this with `ClassDefNotFoundError`s, I wasn't sure how to resolve this without just adding a test dependency in a bunch of modules on the new `com.fasterxml.jackson.datatype:jackson-datatype-jsr310` dependency. 

I'm also curious to hear if you think this is even a fair expectation of the SDK, or if we should just require customers to cast their timestamps to primitives before adding it as metadata.

### DIFF
--- a/aws-xray-recorder-sdk-core/src/main/java/com/amazonaws/xray/entities/EntityImpl.java
+++ b/aws-xray-recorder-sdk-core/src/main/java/com/amazonaws/xray/entities/EntityImpl.java
@@ -33,7 +33,6 @@ import com.fasterxml.jackson.databind.module.SimpleModule;
 import com.fasterxml.jackson.databind.node.NullNode;
 import com.fasterxml.jackson.databind.ser.BeanSerializerModifier;
 import com.fasterxml.jackson.databind.ser.std.ToStringSerializer;
-import com.fasterxml.jackson.datatype.jsr310.JavaTimeModule;
 import com.google.errorprone.annotations.concurrent.GuardedBy;
 import java.util.ArrayList;
 import java.util.HashMap;
@@ -58,7 +57,7 @@ public abstract class EntityImpl implements Entity {
     @SuppressWarnings("checkstyle:ConstantName")
     @Deprecated
     protected static final ObjectMapper mapper = new ObjectMapper()
-        .registerModule(new JavaTimeModule())
+        .findAndRegisterModules()
         .setPropertyNamingStrategy(PropertyNamingStrategy.CAMEL_CASE_TO_LOWER_CASE_WITH_UNDERSCORES)
         .setSerializationInclusion(JsonInclude.Include.NON_EMPTY);
 
--- a/aws-xray-recorder-sdk-core/src/main/java/com/amazonaws/xray/entities/EntityImpl.java
+++ b/aws-xray-recorder-sdk-core/src/main/java/com/amazonaws/xray/entities/EntityImpl.java
@@ -33,6 +33,7 @@ import com.fasterxml.jackson.databind.module.SimpleModule;
 import com.fasterxml.jackson.databind.node.NullNode;
 import com.fasterxml.jackson.databind.ser.BeanSerializerModifier;
 import com.fasterxml.jackson.databind.ser.std.ToStringSerializer;
+import com.fasterxml.jackson.datatype.jsr310.JavaTimeModule;
 import com.google.errorprone.annotations.concurrent.GuardedBy;
 import java.util.ArrayList;
 import java.util.HashMap;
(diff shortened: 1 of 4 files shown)


aws-xray-recorder-sdk-core/src/main/java/com/amazonaws/xray/entities/EntityImpl.java
```diff
@@ -57,6 +58,7 @@
     @SuppressWarnings("checkstyle:ConstantName")
     @Deprecated
     protected static final ObjectMapper mapper = new ObjectMapper()
+        .registerModule(new JavaTimeModule())
```
anuraaga, line 61: I suppose we can use the autoFindModule thing that object mapper has to deal with this situation.

anuraaga: Can you confirm the values are formatted correctly? I think JavaTimeModule uses ISO format IIRC, which I don't think we want in x-ray model.

anuraaga (commented):
It would definitely be nice if our API had strict types like OTel but given we accept `Object` everywhere the expectations are challenging. That being said, we can't support any possible type naturally. Supporting these sort of primitives seems ok to me though, we need to add unit tests probably using test-sets to add the dependency.

willarmiros:
@anuraaga addressed comments

aws-xray-recorder-sdk-core/src/test/java/com/amazonaws/xray/entities/EntityTest.java
```diff
@@ -0,0 +1,21 @@
...
+import com.amazonaws.xray.AWSXRay;
+import java.time.Duration;
+import org.junit.jupiter.api.Test;
+
+class EntityTest {
+
+    @Test
+    void testTimeModuleSerialization() {
+        Segment seg = new SegmentImpl(AWSXRay.getGlobalRecorder(),"test");
+        Duration duration = Duration.ofSeconds(1);
```
anuraaga, line 14: Can you do some test cases for smaller and larger values too? And also timestamps?

willarmiros: Added!

aws-xray-recorder-sdk-core/src/test/java/com/amazonaws/xray/entities/EntityTest.java
```diff
@@ -0,0 +1,56 @@
...
+        String serializedSeg = seg.serialize();
+
+        String expected = "{\"default\":{\"millisecond\":0.003000000,\"second\":1.000000000,\"minute\":3300.000000000}}";
+        assertThat(serializedSeg).contains(expected);
+    }
+
+    @Test
+    void testTimestampSerialization() {
+        Segment seg = new SegmentImpl(AWSXRay.getGlobalRecorder(), "test");
+        seg.putMetadata("date", Date.from(Instant.ofEpochSecond(1616559298)));
```
anuraaga, line 50: Shouldn't the module allow serializing Instant without converting to Date?

aws-xray-recorder-sdk-core/src/test/java/com/amazonaws/xray/entities/EntityTest.java
```diff
@@ -0,0 +1,56 @@
...
+        String serializedSeg = seg.serialize();
+
+        String expected = "{\"default\":{\"date\":1616559298000}}";
```
anuraaga, line 53: I guess this is a millisecond int but usually we serialize times as second double. Is this format ok?

willarmiros: Huh yeah `Instant` serializes to a seconds double but `Date` doesn't. All I found was the default behavior of Jackson is to serialize Dates as timestamps rather than ISO strings, but didn't explain why they are `int` instead of `double`: https://fasterxml.github.io/jackson-databind/javadoc/2.6/com/fasterxml/jackson/databind/SerializationFeature.html#WRITE_DATES_AS_TIMESTAMPS

I guess maybe because Dates are only accurate to the nearest millisecond? First line of Javadoc: https://docs.oracle.com/javase/8/docs/api/java/util/Date.html

aws-xray-recorder-sdk-core/src/test/java/com/amazonaws/xray/entities/EntityTest.java
willarmiros: Yeah, wasn't thinking. Changed to Instant

aws-xray-recorder-sdk-core/src/test/java/com/amazonaws/xray/entities/EntityTest.java
willarmiros: I can add a separate test for Date if you want, to indicate the int format is expected

anuraaga: Yeah let's go ahead and add a test for Date as well at least for documentation. This PR didn't change it's behavior anyways so it's ok.

anuraaga (approved):
Just small comment for one more test

PR #270, opened 2021-03-26
Adds enhanced metadata parsing for Oracle DBs

willarmiros:
*Description of changes:*
This change enables users of the X-Ray SDK and agent who query Oracle DBs to have more accurate metadata by directly parsing the DB connection URL. Much of the parsing logic is adapted from [OpenTelemetry](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/28b1aa386bde65647a95d9e7d435a0894d52cfa1/instrumentation/jdbc/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/jdbc/JdbcConnectionUrlParser.java), and attributed as such. 

Since we are likely not going to add this advanced support for other DB providers, I mostly special-cased the Oracle logic as opposed to the more generic approach that OTel takes for simplicity. I always preferred the info from parsing over the info from `DatabaseMetaData` where applicable.

This also fixes a more general issue with DB subsegment naming where we didn't do sufficient null-checking.

### DIFF
--- a/aws-xray-recorder-sdk-sql/src/main/java/com/amazonaws/xray/sql/ConnectionInfo.java
+++ b/aws-xray-recorder-sdk-sql/src/main/java/com/amazonaws/xray/sql/ConnectionInfo.java
@@ -1,0 +1,117 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package com.amazonaws.xray.sql;
+
+import java.util.Objects;
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+final class ConnectionInfo {
+    @Nullable
+    private final String sanitizedUrl;
+    @Nullable
+    private final String user;
+    @Nullable
+    private final String host;
+    @Nullable
+    private final String dbName;
+
+    private ConnectionInfo(@Nullable String sanitizedUrl,
+                           @Nullable String user,
+                           @Nullable String host,
+                           @Nullable String dbName) {
+        this.sanitizedUrl = sanitizedUrl;
+        this.user = user;
+        this.host = host;
+        this.dbName = dbName;
+    }
+
+    String getSanitizedUrl() {
+        return sanitizedUrl;
+    }
+
+    String getUser() {
+        return user;
+    }
+
+    String getHost() {
+        return host;
+    }
+
+    String getDbName() {
+        return dbName;
+    }
+
+    @Override
+    public String toString() {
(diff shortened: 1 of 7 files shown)


aws-xray-recorder-sdk-sql/src/main/java/com/amazonaws/xray/sql/ConnectionInfo.java
```diff
@@ -0,0 +1,107 @@
...
+        ConnectionInfo that = (ConnectionInfo) o;
+        return Objects.equals(sanitizedUrl, that.sanitizedUrl) &&
+            Objects.equals(user, that.user) &&
+            Objects.equals(host, that.host) &&
+            Objects.equals(dbName, that.dbName);
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hashCode(sanitizedUrl);
```
willarmiros, line 74: Is this sufficient to only hash the sanitized URL? Originally I hashed all 4 fields, but checkStyle called me out for using `hash` instead of `hashCode`.

anuraaga: Did you use `Objects.hash(sanitizedUrl, user, host, dbName)` (varargs)? That's pretty weird. Maybe you can use IntelliJ to generate equals / hashcode so it just uses multiply / add directly

aws-xray-recorder-sdk-sql/src/main/java/com/amazonaws/xray/sql/OracleConnectionUrlParser.java
```diff
@@ -0,0 +1,169 @@
...
+package com.amazonaws.xray.sql;
+
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+
+/**
+ * Class for parsing Oracle database connection URLs and extracting useful metadata.
+ * Adapted from the OpenTelemetry JdbcConnectionUrlParser class under the Apache 2.0 license.
```
anuraaga, line 25: For all the classes adapted, can you add a second copyright header with the copy from OTel?

aws-xray-recorder-sdk-sql/src/main/java/com/amazonaws/xray/sql/ConnectionInfo.java
willarmiros: Yeah, `Objects.hash()` is actually what IntelliJ generate used by default.

aws-xray-recorder-sdk-sql/src/main/java/com/amazonaws/xray/sql/OracleConnectionUrlParser.java
willarmiros: Sure, good callout.

aws-xray-recorder-sdk-sql/src/main/java/com/amazonaws/xray/sql/ConnectionInfo.java
anuraaga: I think you can use "IntelliJ default" to get the multiply version

willarmiros: Ohhh I had never seen that before, thanks for the tip!

anuraaga (approved):

PR #272, opened 2021-04-06
added DockerPlugin for log correlation

willarmiros:
*Description of changes:*
Adds a new plugin for Dockerized applications that parses log groups from the Docker daemon config file to record them in traces. Unfortunately, this parsing still requires manual action from the customer: they need to copy over the config file into their application container so it's readable by the plugin. This is because the docker daemon runs on the host, so its config file is also located on the host, which means it can't be read by containers unless the volume containing the file is bound to the container.

Also refactored some of the EC2Plugin logic to be cleaner.

### DIFF
--- a/aws-xray-recorder-sdk-core/src/main/java/com/amazonaws/xray/utils/DockerUtils.java
+++ b/aws-xray-recorder-sdk-core/src/main/java/com/amazonaws/xray/utils/DockerUtils.java
@@ -15,6 +15,10 @@
 
 package com.amazonaws.xray.utils;
 
+import com.amazonaws.xray.entities.AWSLogReference;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.databind.DeserializationFeature;
+import com.fasterxml.jackson.databind.ObjectMapper;
 import java.io.BufferedReader;
 import java.io.File;
 import java.io.FileInputStream;
@@ -25,6 +29,7 @@ import java.net.MalformedURLException;
 import java.net.URISyntaxException;
 import java.net.URL;
 import java.nio.charset.StandardCharsets;
+import java.nio.file.FileSystem;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 import org.checkerframework.checker.nullness.qual.MonotonicNonNull;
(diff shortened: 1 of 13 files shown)


aws-xray-recorder-sdk-core/src/main/java/com/amazonaws/xray/utils/DockerUtils.java
```diff
@@ -96,4 +105,70 @@ public String getContainerId() throws IOException {
...
+     * Daemon config file: https://docs.docker.com/engine/reference/commandline/dockerd/#daemon-configuration-file
+     * awslogs driver format: https://docs.docker.com/config/containers/logging/awslogs/
+     *
+     * @param filePath file system path to daemon configuration file
+     * @param fs file system to search for config file
+     * @return an AWSLogReference for the discovered log group, or null if the config file is not present or does not
+     *         contain an awslogs driver
+     */
+    @Nullable
+    public static AWSLogReference getAwsLogReference(String filePath, FileSystem fs) {
```
anuraaga, line 121: There doesn't seem to be a need to put this here instead of private where it's used

aws-xray-recorder-sdk-core/src/main/java/com/amazonaws/xray/utils/FSUtils.java
```diff
@@ -0,0 +1,45 @@
...
+ * permissions and limitations under the License.
+ */
+
+package com.amazonaws.xray.utils;
+
+import com.amazonaws.xray.entities.StringValidator;
+import java.nio.file.FileSystem;
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+public class FSUtils {
```
anuraaga, line 22: Ditto, let's not add any public API we don't need (or we could put in `internal` package if we really wanted to)

anuraaga (commented):
Hmm - I see that `daemon.json` isn't the only way to configure docker

https://docs.docker.com/config/daemon/#configure-the-docker-daemon

So not only is mounting daemon.conf into the container hacky, possibly insecure?, but also not complete. I think instead we should just read from some environment variables and ask the customer to set them, would that work?

willarmiros:
Per discussion with @anuraaga, we are discarding this PR in favor of #273.

PR #273, opened 2021-04-06
added mechanism to set log group via env var

willarmiros:
*Description of changes:*
Simple addition to allow customers to set a custom value for a log group via the `AWS_LOG_GROUP` environment variable.


aws-xray-recorder-sdk-core/src/main/java/com/amazonaws/xray/AWSXRayRecorderBuilder.java
```diff
@@ -351,6 +353,16 @@ public AWSXRayRecorder build() {
             }
         });
 
+        String logGroupFromEnv = System.getenv(LOG_GROUP_KEY);
+        if (StringValidator.isNotNullOrBlank(logGroupFromEnv)) {
+            logger.info("Recording log group " + logGroupFromEnv + " from environment variable.");
+            AWSLogReference logReference = new AWSLogReference();
+            logReference.setLogGroup(logGroupFromEnv);
+            Set<AWSLogReference> logReferences = new HashSet<>();
```
anuraaga, line 361: Think you can use something like `Collections.singleton`

anuraaga (approved):

PR #288, opened 2021-06-16
add retry in smoke test

wangzlei:
*Issue #, if available:*
Github action might have random dependency downloading failure. Add retry to reduce false alarmm

*Description of changes:*

### DIFF
--- a/.github/workflows/continuous-monitoring.yml
+++ b/.github/workflows/continuous-monitoring.yml
@@ -32,11 +32,11 @@ jobs:
           
       - name: Run smoke tests
         id: distribution-availability
-        uses: nick-invision/retry@v2
-        with:
-          timeout_seconds: 180
-          max_attempts: 3
-          command: ./gradlew :smoke-tests:check -PtestDistributionChannel=true
+        run: |
+          (echo "===== Attempt: 1 ====" && ./gradlew :smoke-tests:check -PtestDistributionChannel=true) || \
+          (echo "===== Attempt: 2 ====" && ./gradlew :smoke-tests:check -PtestDistributionChannel=true) || \
+          (echo "===== Attempt: 3 ====" && ./gradlew :smoke-tests:check -PtestDistributionChannel=true) || \
+          (echo "===== Failed ====" && exit 1)
         
       - name: Publish metric on X-Ray Java SDK distribution availability
         if: ${{ always() }}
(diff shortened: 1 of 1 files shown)


wangzlei:
aws-otel-lambda canary failure has been solved by [adding retry](https://github.com/aws-observability/aws-otel-lambda/commit/ce758a9b689b6771df65e25d7e8b8be6ccfb0fba#diff-4b56c7c159ef6d182d11a4f6d189ea69d9ee8b2dfde7e7e920052190549c65e8)
https://github.com/aws-observability/aws-otel-lambda/actions/workflows/canary.yml

.github/workflows/continuous-monitoring.yml
```diff
@@ -32,7 +32,11 @@ jobs:
           
       - name: Run smoke tests
         id: distribution-availability
-        run: ./gradlew :smoke-tests:check -PtestDistributionChannel=true
+        uses: nick-invision/retry@v2
```
anuraaga, line 35: We should avoid random third party actions if possible, especially in a distribution channel monitor. I think we can update the cache action to cache dependencies as well instead

wangzlei: Github action may occur downloading failure in anywhere, not only gradle.zip, if we cannot predict which part would get failure it is hard to add cache, now we just see an error then fix it. Also, adding more cache will inflate CI/CD code.

When we testing if the target binary/dependency is available, say xray-java-sdk in this case, we cannot add cache for it. Retry might be the only way to mitigate false download failure.

.github/workflows/continuous-monitoring.yml
```diff
@@ -32,7 +32,11 @@ jobs:
           
       - name: Run smoke tests
         id: distribution-availability
-        run: ./gradlew :smoke-tests:check -PtestDistributionChannel=true
+        run: |
+          (echo "===== Attempt: 1 ====" && ./gradlew :smoke-tests:check -PtestDistributionChannel=true) || \
+          (echo "===== Attempt: 2 ====" && ./gradlew :smoke-tests:check -PtestDistributionChannel=true) || \
+          (echo "===== Attempt: 3 ====" && ./gradlew :smoke-tests:check -PtestDistributionChannel=true) || \
+          (echo "===== Failed ====" && exit 1)
```
NathanielRN, line 39:
```suggestion
          (echo "===== No more retries. Failed. ====" && exit 1)
```

NathanielRN (approved):
Very slick use of shell commands :)

anuraaga (approved):

PR #294, opened 2021-08-02
Improvements to no-op subsegment behavior

willarmiros:
*Description of changes:*
This PR includes a few changes involving no-op subsegment behavior to correct regressions:
1. Add a new `shouldPropagate` field to subsegments, so that we can distinguish between no-op subsegments that are unsampled and no-op subsegments which are created from context missing exceptions.
2. In LambdaSegmentContext, fixed the `endSubsegment` behavior to use the context missing strategy instead of always throwing an exception.
3. Changed the logging logic for ending no-op subsegments to make debug logs more clear

### DIFF
--- a/aws-xray-recorder-sdk-core/src/main/java/com/amazonaws/xray/entities/Subsegment.java
+++ b/aws-xray-recorder-sdk-core/src/main/java/com/amazonaws/xray/entities/Subsegment.java
@@ -85,12 +85,6 @@ public interface Subsegment extends Entity {
     boolean shouldPropagate();
 
     /**
-     * Determines if this subsegment should propagate its trace context downstream
-     * @param shouldPropagate
-     */
-    void setShouldPropagate(boolean shouldPropagate);
-
-    /**
      * Serializes the subsegment as a standalone String with enough information for the subsegment to be streamed on its own.
      * @return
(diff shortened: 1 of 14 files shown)


aws-xray-recorder-sdk-core/src/main/java/com/amazonaws/xray/entities/Subsegment.java
```diff
@@ -78,6 +78,18 @@ static Subsegment noOp(Segment parent, AWSXRayRecorder recorder) {
...
+     * Determines if this subsegment should propagate its trace context downstream
+     * @return true if its trace context should be propagated downstream, false otherwise
+     */
+    boolean shouldPropagate();
+
+    /**
+     * Determines if this subsegment should propagate its trace context downstream
+     * @param shouldPropagate
+     */
+    void setShouldPropagate(boolean shouldPropagate);
```
anuraaga, line 91: Don't think we need this setter

willarmiros: Fair, will make it `final`

willarmiros:
@anuraaga any chance for a second look here and on #293?

anuraaga (approved):

PR #299, opened 2021-09-01
Remove retry count assertions from unit tests

srprash:
The number of retries on a failed request is Lambda dependent in these test cases and may break the assertions if the retry counts are changed in the dependency. Since it is not in the scope of the code being tested, we can safely drop these assertions.


willarmiros (approved):
Thanks!

Issue #303, opened 2021-09-29
Code deadlock due to synchronized blocks


narras-oss:
We noticed a code deadlock within few hours f upgrading to Xray SDK version 2.9.1 from 2.8.0 at these stack trace locations.
We reverted back to 2.8.0 since the synchronized() blocks weren't there in EntityImpl in version 2.8.0.

Thread 1:
java.lang.Thread.State: BLOCKED (on object monitor)
at com.amazonaws.xray.entities.SegmentImpl.isSampled(SegmentImpl.java:91)
- waiting to lock <0x000000008527a6b8> (a java.lang.Object)
at com.amazonaws.xray.strategy.DefaultStreamingStrategy.requiresStreaming(DefaultStreamingStrategy.java:70)
at com.amazonaws.xray.contexts.ThreadLocalSegmentContext.endSubsegment(ThreadLocalSegmentContext.java:81)
at com.amazonaws.xray.AWSXRayRecorder.endSubsegment(AWSXRayRecorder.java:628)
at com.amazonaws.xray.entities.SubsegmentImpl.close(SubsegmentImpl.java:167)
- locked <0x000000008527a6a8> (a java.lang.Object)

Thread2:
java.lang.Thread.State: BLOCKED (on object monitor)
at com.amazonaws.xray.entities.EntityImpl.getId(EntityImpl.java:237)
- waiting to lock <0x000000008527a6a8> (a java.lang.Object)
at jdk.internal.reflect.GeneratedMethodAccessor201.invoke(Unknown Source)
at jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(java.base@11.0.12/Unknown Source)
at java.lang.reflect.Method.invoke(java.base@11.0.12/Unknown Source)
at com.fasterxml.jackson.databind.ser.BeanPropertyWriter.serializeAsField(BeanPropertyWriter.java:689)
at com.fasterxml.jackson.databind.ser.std.BeanSerializerBase.serializeFields(BeanSerializerBase.java:770)
at com.fasterxml.jackson.databind.ser.BeanSerializer.serialize(BeanSerializer.java:178)
at com.fasterxml.jackson.databind.ser.impl.IndexedListSerializer.serializeContents(IndexedListSerializer.java:119)
at com.fasterxml.jackson.databind.ser.impl.IndexedListSerializer.serialize(IndexedListSerializer.java:79)
[...]

PR #306, opened 2021-09-30
Don't lock segment from subsegment

anuraaga:
*Issue #, if available:* #303

*Description of changes:*

High mutability and circular refrences between segments and subsegments create a prime landscape for thread issues. This tries to solve #303 by removing locking on the codepaths from `DefaultStreamingStrategy` - the streaming strategy is used when processing subsegments which may be on a different thread than the segment, which could then try to stream subsegments.

This should avoid deadlocks, I don't know if it's possible to be totally confident without removing streaming strategy completely - alway streaming, removing the circular references between segments / subsegments completely, would give confidence they can't deadlock. Probably best to shift to OTel instead though.

### DIFF
--- a/aws-xray-recorder-sdk-core/src/main/java/com/amazonaws/xray/entities/EntityImpl.java
+++ b/aws-xray-recorder-sdk-core/src/main/java/com/amazonaws/xray/entities/EntityImpl.java
@@ -36,10 +36,10 @@ import com.fasterxml.jackson.databind.ser.BeanSerializerModifier;
 import com.fasterxml.jackson.databind.ser.std.ToStringSerializer;
 import com.google.errorprone.annotations.concurrent.GuardedBy;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.concurrent.atomic.LongAdder;
 import java.util.concurrent.locks.ReentrantLock;
 import org.apache.commons.logging.Log;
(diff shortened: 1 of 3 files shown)


aws-xray-recorder-sdk-core/src/main/java/com/amazonaws/xray/entities/EntityImpl.java
```diff
@@ -197,7 +198,7 @@ protected EntityImpl(AWSXRayRecorder creator, String name) {
 
         this.creator = creator;
         this.name = name;
-        this.subsegments = new ArrayList<>();
+        this.subsegments = new CopyOnWriteArrayList<>();
```
willarmiros, line 201: This seems like it could get very expensive for people adding lots of subsegments... I know we don't want to spend the time for the most performant solution, but this seems like a pretty substantial degradation that would impact nearly all customers for more of an edge-case deadlock. Is there any other workaround for this?

anuraaga: I think for something fairly standard like 5 subsegments, this is probably faster than something like a synchronized list.

For a hundred segments it does look slow - but so does the normal `ArrayList` since we call `remove` on it which results in quadratic behavior anyways.

I think to be optimal we would need to use a concurrent hash set, but unfortunately we need to support `getSubsegments().add()`. It is possible to wrap a set into a list (with slow performances of `get(I)`) I think, it could be an idea for if there are reports of regression in the wild.

willarmiros: Ok gotcha, I guess we're also still calling `remove` so there's going to definitely be some degradation there. Can we add a quick additional benchmark test for creating say 100 subsegments just to confirm it's not breaking anything?

PR #309, opened 2021-11-11
Reduce log level of sampling error logging

willarmiros:
*Issue #, if available:*
https://github.com/aws/aws-xray-java-agent/issues/65

*Description of changes:*
Given that errors retrieving sampling rules/targets are non-impacting to the fundamental functionality of tracing, we shouldn't be logging them at error level, especially since they are typically the result of transient network blips. Customers have complained about the verbose, scary looking error message in their logs.

@anuraaga would appreciate if you took a look.

### DIFF
--- a/aws-xray-recorder-sdk-core/src/main/java/com/amazonaws/xray/strategy/sampling/pollers/RulePoller.java
+++ b/aws-xray-recorder-sdk-core/src/main/java/com/amazonaws/xray/strategy/sampling/pollers/RulePoller.java
@@ -71,7 +71,7 @@ public class RulePoller {
             try {
                 pollRule();
             } catch (Throwable t) {
-                logger.warn("Encountered error polling GetSamplingRules: ", t);
+                logger.info("Encountered error polling GetSamplingRules: ", t);
                 // Propagate if Error so executor stops executing.
                 // TODO(anuraaga): Many Errors aren't fatal, this should probably be more restricted, e.g.
                 // https://github.com/openzipkin/brave/blob/master/brave/src/main/java/brave/internal/Throwables.java
--- a/aws-xray-recorder-sdk-core/src/main/java/com/amazonaws/xray/strategy/sampling/pollers/RulePoller.java
+++ b/aws-xray-recorder-sdk-core/src/main/java/com/amazonaws/xray/strategy/sampling/pollers/RulePoller.java
@@ -71,7 +71,7 @@ public class RulePoller {
             try {
                 pollRule();
             } catch (Throwable t) {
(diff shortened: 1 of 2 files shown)


aws-xray-recorder-sdk-core/src/main/java/com/amazonaws/xray/strategy/sampling/pollers/RulePoller.java
```diff
@@ -71,7 +71,7 @@ public void start() {
             try {
                 pollRule();
             } catch (Throwable t) {
-                logger.error("Encountered error polling GetSamplingRules: ", t);
+                logger.warn("Encountered error polling GetSamplingRules: ", t);
```
anuraaga, line 74: I think in practice error and warn are the same, probably should switch to info.

willarmiros: The key difference is that warn is not emitted at the SDK's default log level

anuraaga: I don't think SDKs ever have a default log level - it's in the end user app usually in a logback.xml. It is fairly standard practice to only log things above info level I think. This is because many libraries just log all failures as warning and don't use error at all. I guess switching to warn doesn't practically change the chance of being annoyed by the log mmessages for many apps

willarmiros: Ok that makes sense - I guess I'm getting my languages confused with default log levels. Went ahead and changed to info.

anuraaga (approved):

aws-xray-recorder-sdk-core/src/main/java/com/amazonaws/xray/strategy/sampling/pollers/RulePoller.java
yijiang-song: hi! willarmiros! When will the latest version of the log level reduction for sampling error logs be released

willarmiros: @yijiang-song unfortunately we cannot provide ETAs for our releases, please feel free to watch the repo for the next release.

yijiang-song: Thank you! I will continue to pay attention to this issue.

PR #310, opened 2021-11-15
Do not create a new emitter when a shared emitter is provided

hden:
*Issue #, if available:*

Currently AWSXRayRecorder's constructor will create a new emitter even when a shared emitter is specified in the AWSXRayRecorderBuilder. This creates an unnecessary Emitter / Socket for each new AWSXRayRecorder and leads to fire descriptor exhaustion if enough Emitter is created.

*Description of changes:*

Change AWSXRayRecorder's constructor so that it does not create a new emitter when a shared emitter is provided.


aws-xray-recorder-sdk-core/src/main/java/com/amazonaws/xray/AWSXRayRecorder.java
```diff
@@ -138,6 +138,10 @@
     private boolean forcedTraceIdGeneration;
 
     public AWSXRayRecorder() {
+        this(null);
+    }
+
+    public AWSXRayRecorder(Emitter sharedEmitter) {
```
willarmiros, line 144: Not too excited about increasing our API surface for this, maybe we can make the new constructor package private?

willarmiros (commented):
Just a quick comment, thanks!

aws-xray-recorder-sdk-core/src/main/java/com/amazonaws/xray/AWSXRayRecorder.java
hden: Sure.

willarmiros:
@hden There's still some syntax errors it seems

hden:
@willarmiros Oops.

hden:
@willarmiros I've fixed the type annotations. Please take a look and tell me what you think.

willarmiros (commented):
Code lgtm, though one thing I'm concerned about in general is thread safety. Having several recorders use the same emitter means this path could be executed concurrently for the same socket: https://github.com/aws/aws-xray-sdk-java/blob/7e935116ad45b109831d4dba6f178f4d9c22ce8b/aws-xray-recorder-sdk-core/src/main/java/com/amazonaws/xray/emitters/UDPEmitter.java#L111-L124

Is there any possibility of interference, overload, or data corruption for all that traffic going to a socket at once? Also if I may ask, what is your use case that requires several recorders which share an emitter? Generally speaking the recorder is meant to be a singleton, with only one per application.

hden:
We've been testing with thread safety as well, so far we have not seen any evidence of interference, overload, or data corruption.

Our use-case is exactly thread safety. We have been handling concurrent workload on a thread-pool (in Clojure). We started with a singleton, but it's quite unrealistic to keep track with segment context since context switching is everywhere. So we've been trying with the idea that we could create one recorder per job (a request, to be precise) and pass the recorder along the job context.

cc @popoppo What do you think?

PR #329, opened 2022-04-04
updated spring dependencies

bhautikpip:
*Issue #, if available:*
Spring RCE issue: https://tanzu.vmware.com/security/cve-2022-22965

*Description of changes:*
Updated spring dependencies

### DIFF
--- a/aws-xray-recorder-sdk-spring/build.gradle.kts
+++ b/aws-xray-recorder-sdk-spring/build.gradle.kts
@@ -9,11 +9,11 @@ dependencies {
     api("org.aspectj:aspectjrt:1.8.11")
 
     // TODO(anuraaga): Remove most of these? Seems only Configurable annotation is used
-    implementation("org.springframework:spring-context-support:4.3.12.RELEASE")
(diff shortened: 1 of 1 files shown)


aws-xray-recorder-sdk-spring/build.gradle.kts
```diff
@@ -9,11 +9,11 @@ dependencies {
     api("org.aspectj:aspectjrt:1.8.11")
 
     // TODO(anuraaga): Remove most of these? Seems only Configurable annotation is used
```
srprash, line 11: @anuraaga 's comment hints that these dependencies could be removed. If it is not a big effort, maybe this is a good time to do so?
Otherwise LGTM.

srprash (approved):

aws-xray-recorder-sdk-spring/build.gradle.kts
bhautikpip: Yeah Good callout. I think it's a small change relatively but in order to ensure nothing breaks and SDK works as expected we might need to do some additional testing so I have scoped that work out of this PR if that makes sense.

PR #256, opened 2021-01-28
update aws-java-sdk to something much less ancient

Algae666:
*Issue #, if available:*

*Description of changes:*

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #334, opened 2022-05-25
Upgrade AWS Java SDK to 1.12.228

### LINKED ISSUE #333: Update of dependencies to comply with Whitesource security scan
At our company we're using Whitesource software for detection of the security issues. Unfortunately the newest version (2.11.1) of aws-xray-recorder-sdk-core doesn't comply with our policy. Some libraries have serious vulnerabilities and some are just old (joda-time is 7 years old). Below is the list of libraries that were rejected in a security scan.
Most if not all of these problematic dependencies come from aws-java-sdk-core library which should also be updated. 

image

```
aws-xray-recorder-sdk-core-2.11.1.jar 
└---aws-java-sdk-xray-1.11.903.jar 
       └---aws-java-sdk-core-1.11.903.jar 
                └---ion-java-1.0.2.jar - **_reject old libraries_**
                └---jackson-databind-2.6.7.4.jar - **_reject high vulnerable libraries_**
                        └---jackson-annotations-2.6.0.jar - **_reject old libraries_**
                        └---jackson-core-2.6.7.jar - **_reject old libraries_**
                └---jackson-dataformat-cbor-2.6.7.jar - **_reject high vulnerable libraries_**
                └---joda-time-2.8.1.jar - **_reject old libraries_**
```

Please consider this request as without these changes it becomes impossible to use xray.

NathanielRN:
*Issue #, if available:* #333

*Description of changes:*

Upgrade AWS SDK for Java to version `1.12.228`. This was inspired by the same changes we made in https://github.com/aws/aws-xray-sdk-java/pull/256.

We fixed several issues along the way:
* All of a sudden when we updated to AWS SDK version `1.12.228`, many dependencies could not be resolved even though they were definitely there. Worked with a colleague to find out that [this is a Gradle 6 issue](https://github.com/gradle/gradle/issues/8489#issuecomment-646346668). Updating gralde with `./gradlew wrapper --gradle-version=7.4.2
` fixed the issue 🥳
  * This was the error we were seeing which we were able to Google:
    ```
     What went wrong:
     Execution failed for task ':aws-xray-recorder-sdk-core:dependencies'.
     > Could not resolve all dependencies for configuration ':aws-xray-recorder-sdk-core:testCompileClasspath'.
     > Problems reading data from Binary store in /private/var/folders/5s/kz556x054vz8l7204yh4j9krq6bwmm/T/gradle9062832512095609875.bin offset 29727 exists? true
    ```
  * The reason is:
     > it's an issue with the binary serialization that occurs when the dependency graphs are resolved, and our project triggered one of them when we upgrade to AWS Java SDK 1.12.x for some reason.
* We had to update the `junit-bom` to version `5.8.2` because its dependencies [do not support `junit` version `1.12`](https://stackoverflow.com/a/67566548/7460739) which is another package we pull in
* Once we upgrade Gradle to version 7, a package that checks our license version started failing due to missing annotations. This [was fixed in version `0.16.1` of that package](https://github.com/hierynomus/license-gradle-plugin/issues/191) so we updated it here.

Cool thing I learned:

Your first instinct should be to check if all dependencies resolved when working with gradle.

i.e. `/gradlew :a-x-r-s-c:dependencies`

`a-x-r-s-c` is short for `aws-xray-recorder-sdk-core`.

and then google whatever failure comes out 🙂

Fixes #333

### DIFF
--- a/aws-xray-recorder-sdk-sql/src/main/java/com/amazonaws/xray/sql/SqlSubsegments.java
+++ b/aws-xray-recorder-sdk-sql/src/main/java/com/amazonaws/xray/sql/SqlSubsegments.java
@@ -18,14 +18,12 @@ package com.amazonaws.xray.sql;
 import com.amazonaws.xray.AWSXRay;
 import com.amazonaws.xray.entities.Namespace;
 import com.amazonaws.xray.entities.Subsegment;
-
 import com.blogspot.mydailyjava.weaklockfree.WeakConcurrentMap;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.sql.Connection;
 import java.sql.DatabaseMetaData;
 import java.sql.SQLException;
-
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 import org.checkerframework.checker.nullness.qual.Nullable;
--- a/dependencyManagement/build.gradle.kts
+++ b/dependencyManagement/build.gradle.kts
@@ -5,8 +5,9 @@ plugins {
 data class DependencySet(val group: String, val version: String, val modules: List<String>)
 
 val DEPENDENCY_BOMS = listOf(
-        "com.fasterxml.jackson:jackson-bom:2.11.0",
-        "org.junit:junit-bom:5.6.2"
+        "com.amazonaws:aws-java-sdk-bom:1.12.228",
+        "com.fasterxml.jackson:jackson-bom:2.12.0",
+        "org.junit:junit-bom:5.8.2"
 )
 
 val DEPENDENCY_SETS = listOf(
@@ -17,7 +18,7 @@ val DEPENDENCY_SETS = listOf(
         ),
         DependencySet(
                 "com.fasterxml.jackson.datatype",
-                "2.11.0",
+                "2.12.0",
                 listOf("jackson-datatype-jsr310")
         ),
         DependencySet(
@@ -42,7 +43,7 @@ val DEPENDENCY_SETS = listOf(
         ),
         DependencySet(
                 "junit",
-                "4.12",
+                "4.13.1",
                 listOf("junit")
         ),
         DependencySet(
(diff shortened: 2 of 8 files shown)


aws-xray-recorder-sdk-sql/src/main/java/com/amazonaws/xray/sql/SqlSubsegments.java
```diff
@@ -18,14 +18,12 @@
 import com.amazonaws.xray.AWSXRay;
 import com.amazonaws.xray.entities.Namespace;
 import com.amazonaws.xray.entities.Subsegment;
-
```
NathanielRN, line 21: These formatting changes were failing the build, although I don't agree with them, added them for now and we can revisit the linter later.

dependencyManagement/build.gradle.kts
```diff
@@ -5,8 +5,9 @@ plugins {
 data class DependencySet(val group: String, val version: String, val modules: List<String>)
 
 val DEPENDENCY_BOMS = listOf(
-        "com.fasterxml.jackson:jackson-bom:2.11.0",
-        "org.junit:junit-bom:5.6.2"
+        "com.amazonaws:aws-java-sdk-bom:1.12.228",
+        "com.fasterxml.jackson:jackson-bom:2.12.0",
```
NathanielRN, line 9: I wanted to remove this `jackson` reference and the one below, but it failed to find them at `compileOnly` time. I was hoping it would get it through the `com.amazonaws:aws-java-sdk-bom` dependency which requires them.

However, now my theory is that it won't see them because `com.amazonaws:aws-java-sdk-bom` doesn't import them in a way that is visible to downstream packages. So we will have to continue updating it for future AWS Java SDK updates.

rli: v1.12.x shadows Jackson dependencies since customers were relying on the version that came with the SDK, preventing the team from upgrading from insecure versions to newer ones without introducing breaking changes.

Hence the change in v1.12.x which forces customers to bring their own Jackson version.

willarmiros (approved):
Awesome! Looks like a great fix 🤠

wangzlei (approved):

Issue #318, opened 2022-01-27
Deadlock in DefaultStreamingStrategy when ending a subsegment with many threads on the same Segment


Kurru:
I have some code that generates many subsegments rapidly using multiple threads. I've observed that they can get dead locked on each other due to the EntityImpl.isInProgress lock used in DefaultStreamingStrategy. 

It seems to end a subsegment, the code first takes a lock on the subsegment (`AWSXRayRecorder.endSubsegment(AWSXRayRecorder.java:630`) then they attempt to check there are no locks on any other subsegment within the Segment associated with the subsegment (`ThreadLocalSegmentContext.endSubsegment(ThreadLocalSegmentContext.java:86`) and eventually `DefaultStreamingStrategy.stream(DefaultStreamingStrategy.java:103)`

If 2 different threads attempt to end a subsegment at the same time (for the same segment) AND the segment is large enough to trigger segment streaming, then both threads will lock. In practice, if there are multiple threads working, they will all eventually want to end their segments and thus all outstanding threads interacting with this Segment will also lock waiting for the deadlock to clear.

Thread grpc-default-executor-0
```
"grpc-default-executor-0" prio=0 tid=0x0 nid=0x0 blocked
     java.lang.Thread.State: BLOCKED
 on java.lang.Object@264dd439 owned by "grpc-default-executor-3" Id=74
	at app//com.amazonaws.xray.entities.EntityImpl.isInProgress(EntityImpl.java:483)
	at app//com.amazonaws.xray.strategy.DefaultStreamingStrategy.stream(DefaultStreamingStrategy.java:103)
	at app//com.amazonaws.xray.strategy.DefaultStreamingStrategy.stream(DefaultStreamingStrategy.java:96)
	at app//com.amazonaws.xray.strategy.DefaultStreamingStrategy.stream(DefaultStreamingStrategy.java:96)
	at app//com.amazonaws.xray.strategy.DefaultStreamingStrategy.stream(DefaultStreamingStrategy.java:96)
	at app//com.amazonaws.xray.strategy.DefaultStreamingStrategy.streamSome(DefaultStreamingStrategy.java:86)
	at app//com.amazonaws.xray.contexts.ThreadLocalSegmentContext.endSubsegment(ThreadLocalSegmentContext.java:86)
	at app//com.amazonaws.xray.AWSXRayRecorder.endSubsegment(AWSXRayRecorder.java:630)
[...]

PR #342, opened 2022-07-22
Add javadoc to Subsegment.close() method.

Kurru:
*Issue #, if available:*
#318

*Description of changes:*
Add javadoc to Subsegment.close() method.
This method should not be used to end subsegments for segments using multiple threads as this can result in a deadlock. 

### DIFF
--- a/aws-xray-recorder-sdk-core/src/main/java/com/amazonaws/xray/entities/Subsegment.java
+++ b/aws-xray-recorder-sdk-core/src/main/java/com/amazonaws/xray/entities/Subsegment.java
@@ -87,7 +87,7 @@ public interface Subsegment extends Entity {
     /**
      * Serializes the subsegment as a standalone String with enough information for the subsegment to be streamed on its own.
      * @return
-     *  the string representation of the subsegment with enouogh information for it to be streamed
+     *  the string representation of the subsegment with enough information for it to be streamed
      */
     String streamSerialize();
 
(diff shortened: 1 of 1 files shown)


jj22ee:
Hi @Kurru, could you change your javadoc as the .close() method is not designed for external use and should not be used in general, rather than for a specific use case?

Kurru:
I have updated the javadoc as requested. Let me know if this new wording is insufficient.

aws-xray-recorder-sdk-core/src/main/java/com/amazonaws/xray/entities/Subsegment.java
```diff
@@ -96,10 +96,16 @@ static Subsegment noOp(Segment parent, AWSXRayRecorder recorder) {
...
      * @return
-     *  the pretty string representation of the subsegment with enouogh information for it to be streamed
+     *  the pretty string representation of the subsegment with enough information for it to be streamed
      */
     String prettyStreamSerialize();
 
+    /**
+     * Implements the {@link AutoCloseable} interface.
+     *
+     * Warning. This method is intended to private to the xray library and should not be used externally. Instead, use
```
jj22ee, line 106: Small nitpick - This method is intended _be_ to private

Kurru: Done

jj22ee (approved):

PR #345, opened 2022-07-28
Add @Nullable to AWSXray.setTraceEntity() parameter

Kurru:
*Issue #, if available:*
N/A

*Description of changes:*
Add @Nullable to AWSXray.setTraceEntity() parameter as the delegate method in AWSXrayRecorder accepts nulls when clearing the TraceEntity state. 

This will resolve a 'null check' warning in IntelliJ `Argument 'original' might be null`.


jj22ee (approved):
LGTM

PR #361, opened 2022-11-03
Refactoring TraceHeader Calls

atshaw43:
*Issue #, if available:*

*Description of changes:*
**Problem 1**
We have redundant code for creating a trace header.

**Solution**
Create a centralized function for creating trace headers from entities. This required NoOpSubsegment to have its traceId.

**Problem 2**
Why does NoOpSubsegment need a namespace and name? We have a bug where we are not correctly ending subsegments if they are NoOp. The subsegment gets created, but never ended by afterResponse because it is seen as a duplicate because its namespace and name is always empty.

**Solution**
Pass down the name and namespace in NoOpSubsegments.

https://github.com/aws/aws-xray-sdk-java/blob/master/aws-xray-recorder-sdk-aws-sdk/src/main/java/com/amazonaws/xray/handlers/TracingHandler.java

```
@Override
    public void afterResponse(Request<?> request, Response<?> response) {
        if (isSubsegmentDuplicate(recorder.getCurrentSubsegmentOptional(), request)) {
            Optional<Subsegment> currentSubsegmentOptional = recorder.getCurrentSubsegmentOptional();
            if (!currentSubsegmentOptional.isPresent()) {
                return;
            }
            Subsegment currentSubsegment = currentSubsegmentOptional.get();
            populateAndEndSubsegment(currentSubsegment, request, response, null);
        }
    }
```


atshaw43:
Problem: Why does NoOpSubsegment need a namespace? We have a bug where we are not correctly ending subsegments if they are NoOp. The subsegment gets created, but never ended by afterResponse because it is seen as a duplicate because its namespace is always empty.

Solution: Pass down the namespace in NoOpSubsegments.

https://github.com/aws/aws-xray-sdk-java/blob/master/aws-xray-recorder-sdk-aws-sdk/src/main/java/com/amazonaws/xray/handlers/TracingHandler.java

@Override
    public void afterResponse(Request<?> request, Response<?> response) {
        if (isSubsegmentDuplicate(recorder.getCurrentSubsegmentOptional(), request)) {
            Optional<Subsegment> currentSubsegmentOptional = recorder.getCurrentSubsegmentOptional();
            if (!currentSubsegmentOptional.isPresent()) {
                return;
            }
            Subsegment currentSubsegment = currentSubsegmentOptional.get();
            populateAndEndSubsegment(currentSubsegment, request, response, null);
        }
    }

aws-xray-recorder-sdk-aws-sdk/src/main/java/com/amazonaws/xray/handlers/TracingHandler.java
```diff
@@ -194,15 +193,8 @@ public void beforeRequest(Request<?> request) {
...
-            boolean isSampled = previousSubsegment.isPresent() ?
-                    previousSubsegment.get().isSampled() :
-                    recorder.getCurrentSegment().isSampled();
-
-            TraceHeader header =
-                new TraceHeader(recorder.getCurrentSegment().getTraceId(),
-                        isSampled ? currentSubsegment.getId() : null,
-                        isSampled ? SampleDecision.SAMPLED : SampleDecision.NOT_SAMPLED);
-            request.addHeader(TraceHeader.HEADER_KEY, header.toString());
+            request.addHeader(TraceHeader.HEADER_KEY, TraceHeader.fromEntity(
```
srprash, line 196: This code doesn't look equivalent to the one it's replacing.
In the left side code, the parent id as the TraceHeader argument is either `currentSubsegment`'s id or `null`. Only the `isSampled` is fetched from either previousSubsegment or the `recorder.getCurrentSegment()`.
In the right side code, the whole TraceHeader will be created either from `previousSubsegment` or `currentSubsegment`.

aws-xray-recorder-sdk-core/src/main/java/com/amazonaws/xray/entities/NoOpSubSegment.java
```diff
@@ -65,12 +79,14 @@ class NoOpSubSegment implements Subsegment {
...
 
-        this.isSampled = samplingStrategyOverride == SamplingStrategyOverride.DISABLED ?
-                parentSegment.isSampled() :
-                false;
-        this.isRecording = isSampled;
+        this.isSampled = false;
+        this.isRecording = false;
         this.samplingStrategyOverride = samplingStrategyOverride;
+
+        id = creator.getIdGenerator().newEntityId();
```
srprash, line 89: If this `NoOpSubsegment` is never emitted and never propagated, can we set a static id here? We should avoid unnecessary computation if we can.

aws-xray-recorder-sdk-core/src/main/java/com/amazonaws/xray/entities/Segment.java
```diff
@@ -43,6 +43,8 @@ static Segment noOp(TraceID traceId, AWSXRayRecorder recorder) {
     /**
      * @return the sampled
      */
+    @JsonIgnore
+    @Override
```
srprash, line 47: Do we need to necessarily override the `isSampled()` method from `Entity`?
Is it possible for `Entity` types and subtypes to directly use this method from `Entity` without an override?

aws-xray-recorder-sdk-aws-sdk/src/main/java/com/amazonaws/xray/handlers/TracingHandler.java
atshaw43: Wow, good catch! This was a bug.
I looked at the original code and I don't think we even need to use the previous subsegment here.

srprash (approved):

PR #363, opened 2022-11-17
Update README.md

atshaw43:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/README.md
+++ b/README.md
@@ -240,26 +240,25 @@ public class Handler implements RequestHandler<SQSEvent, String> {
   @Override
   public String handleRequest(SQSEvent event, Context context) {
 
-    // Check to see if any messages upstream are sampled.
-    boolean toSample = false;
+    int i = 1;
+
     for (SQSMessage message: event.getRecords()) {
+
+      // Check if the message is sampled
       if (SQSMessageHelper.isSampled(message)) {
-        toSample = true;
+        AWSXRay.beginSubsegment("Processing Message - " + i);
+      } else {
+        AWSXRay.beginSubsegmentWithoutSampling("Processing Message - " + i);
       }
-    }
 
-    // Create a new subsegment
-    if (toSample) {
-      AWSXRay.beginSubsegment("Processing Message");
-    } else {
-      AWSXRay.beginSubsegmentWithoutSampling("Processing Message");
-    }
-    
-    // Do your procesing work here
-    System.out.println("Doing processing work");
+      i++;
 
-    // End your subsegment
-    AWSXRay.endSubsegment();
+      // Do your procesing work here
+      System.out.println("Doing processing work");
+
+      // End your subsegment
+      AWSXRay.endSubsegment();
+    }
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -228,6 +228,43 @@ try {
...
+  }
+
+  @Override
+  public String handleRequest(SQSEvent event, Context context) {
+
+    // Check to see if any messages upstream are sampled.
+    boolean toSample = false;
+    for (SQSMessage message: event.getRecords()) {
+      if (SQSMessageHelper.isSampled(message)) {
+        toSample = true;
```
srprash, line 247: Wouldn't this always sample the subsegment even when a single message in the batch is sampled?
I thought how oversampling mitigation would work is that when processing each message in the batch, we decided whether to begin a subsegment with/without sampling based on if the message is sampled or not.

README.md
```diff
@@ -228,6 +228,43 @@ try {
...
+      AWSXRay.beginSubsegmentWithoutSampling("Processing Message");
+    }
+    
+    // Do your procesing work here
+    System.out.println("Doing processing work");
+
+    // End your subsegment
+    AWSXRay.endSubsegment();
+
+    return "Success";
```
srprash, line 264: If we are processing each message in the batch, shouldn't this code be inside the for loop?

README.md
```diff
@@ -228,6 +228,43 @@ try {
 ```
 Note that in the closure-based example above, exceptions are intercepted automatically.
 
+### Oversampling Mitigation
+Oversampling mitigation allows you to ignore a parent segment/subsegment's sampled flag and instead set it to false.
+The code below demonstrates overriding the sampled flag based on the SQS messages sent to Lambda.
```
srprash, line 233: I think this can be reworded to explain what is happening a little more clearly. 

"The code below demonstrates creating a sampled or un-sampled subsegment based on the sampling status of each SQS message processed by Lambda"
^^ How does something like this sound?

carolabadeer: +1

README.md
atshaw43: I wanted to demonstrate how to do a fan-in.

README.md
atshaw43: See comment above.

README.md
atshaw43: I want to show how to do a fan-in. I don't think the rewording reflects that.

srprash (approved):
LGTM!

README.md
```diff
@@ -228,6 +228,70 @@ try {
 ```
 Note that in the closure-based example above, exceptions are intercepted automatically.
 
+### Oversampling Mitigation
+Oversampling mitigation allows you to ignore a parent segment/subsegment's sampled flag and instead set it to false.
```
carolabadeer, line 232: I think this sentence still needs some rewording - it is unclear what "it" is referring to here

carolabadeer (approved):
LGTM

PR #375, opened 2023-03-14
No class def found error

atshaw43:
*Issue #, if available:*

*Description of changes:*
If a module is needed to serialize an object in Jackson, but it does not exist, then it causes a NoClassDefFoundError

This change will log that error and log which modules are loaded.


srprash (approved):
LGTM

PR #383, opened 2023-05-15
Thread safety

atshaw43:
*Issue #, if available:*
It is possible to get a concurrent modification exception from adding PrecursorIds in two separate threads at the same time.

*Description of changes:*
Make the data structure for PrecursorIds thread safe.

### DIFF
--- a/aws-xray-recorder-sdk-core/src/test/java/com/amazonaws/xray/ConcurrencyTest.java
+++ b/aws-xray-recorder-sdk-core/src/test/java/com/amazonaws/xray/ConcurrencyTest.java
@@ -20,10 +20,8 @@ import com.amazonaws.xray.entities.Segment;
 import com.amazonaws.xray.entities.SegmentImpl;
 import com.amazonaws.xray.entities.SubsegmentImpl;
 import com.amazonaws.xray.strategy.sampling.LocalizedSamplingStrategy;
-
 import java.util.Iterator;
 import java.util.Random;
-import java.util.Vector;
 import java.util.concurrent.Callable;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ExecutorService;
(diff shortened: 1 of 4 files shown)


aws-xray-recorder-sdk-core/src/test/java/com/amazonaws/xray/entities/EntityTest.java
```diff
@@ -76,6 +76,18 @@ void testUnknownClassSerialization() {
...
+        Segment seg = new SegmentImpl(AWSXRay.getGlobalRecorder(), "test");
+        SubsegmentImpl subseg = new SubsegmentImpl(AWSXRay.getGlobalRecorder(), "test", seg);
+        subseg.addPrecursorId("myId1");
+        subseg.addPrecursorId("myId2");
+        String serializedSubSeg = subseg.serialize();
+
+        String expected = "\"precursor_ids\":[\"myId1\",\"myId2\"]";
+        assertThat(serializedSubSeg).contains(expected);
+    }
+
```
carolabadeer, line 90: Is there any way to mock the behaviour of adding precursor IDs in 2 threads and ensure that the new data structure handles that situation correctly?

aws-xray-recorder-sdk-core/src/test/java/com/amazonaws/xray/ConcurrencyTest.java
```diff
@@ -64,4 +70,31 @@ void testManyThreads() throws InterruptedException {
...
+        Thread thread1 = new Thread(() -> {
+            while (true) {
+                subseg.addPrecursorId("ID" + new Random().nextInt());
+            }
+        });
+        thread1.start();
+
+        Callable<Void> callable = (Callable) () -> {
+            while (System.currentTimeMillis() - startTime < TimeUnit.SECONDS.toMillis(1)) {
+                Iterator it = subseg.getPrecursorIds().iterator();
```
jj22ee, line 89: I think this `Callable` should also be using `addPrecursorId` so this function is tested being called in 2 threads.

atshaw43: I was able to produce a concurrent modification exception with this code.
Iterating and changing the data structure is not allowed in a normal HashSet

If they are both adding, then it has to be the same bucket in the hash table at the same time which is harder to get to occur.

jj22ee (approved):
LGTM

PR #387, opened 2023-07-14
Optimize default resolver chain construction

rohitnair:
We can figure out the right SegmentContextResolver at init time since it should remain static throughout the lifetime of an application (we're either running within Lambda or we're not - it should never change). Trying to resolve through the resolver chain at runtime seems unnecessary. 

While typically it should not be very expensive to resolve both resolvers at runtime, I've noticed some performance overhead in environments where JSM (Java Security Manager) is enabled since the System.getenv() call that the LambdaSegmentContextResolver makes internally requires a permissions check (https://github.com/openjdk/jdk/blob/jdk-11%2B28/src/java.base/share/classes/java/lang/System.java#L991-L994) 

*Issue #, if available:*
N/A

*Description of changes:*
See above


atshaw43:
I fixed the tests.
The Lambda environment variables were not set at the time the X-Ray Recorder was created so it defaulted to thread context.

atshaw43 (approved):

PR #393, opened 2023-12-12
Fxing S3 bucket name extraction for AWS SDK V2

atshaw43:
*Issue #, if available:*
In AWS SDK V2 the S3 bucket field name was changed from BucketName to Bucket. This caused us to no longer add the field to the AWS field in the segment. Which caused it to not show up in the console.

*Description of changes:*
I changed it so the AWS SDK V2 instrumentation picks up the new field. This changed the field name going to the X-Ray backend, so I also added a piece of code that changes the field name back to "bucket_name" for them.


wangzlei (approved):
LGTM!

PR #403, opened 2024-05-06
[Lambda] Send NoOp segment when trace header is incomplete

majanjua-amzn:
When using Lambda, if there is no trace header or an invalid one we would like to send a NoOp segment instead of a Facade segment so that customers can call X-Ray SDK APIs and utilize X-Ray SDK without any output. This will not send data to X-Ray nor inject X-Ray trace headers downstream.

*Description of changes:*
- Modify condition to create a NoOp segment when trace header is "initializing", i.e. it is incomplete
- Ensure trace entity is cleared when ending the segment
- Make `NoOpSegment` public for the above changes
- Update unit tests

### DIFF
--- a/aws-xray-recorder-sdk-core/src/main/java/com/amazonaws/xray/contexts/LambdaSegmentContext.java
+++ b/aws-xray-recorder-sdk-core/src/main/java/com/amazonaws/xray/contexts/LambdaSegmentContext.java
@@ -23,7 +23,6 @@ import com.amazonaws.xray.entities.Segment;
 import com.amazonaws.xray.entities.Subsegment;
 import com.amazonaws.xray.entities.SubsegmentImpl;
 import com.amazonaws.xray.entities.TraceHeader;
-import com.amazonaws.xray.entities.TraceHeader.SampleDecision;
 import com.amazonaws.xray.entities.TraceID;
 import com.amazonaws.xray.exceptions.SubsegmentNotFoundException;
 import com.amazonaws.xray.listeners.SegmentListener;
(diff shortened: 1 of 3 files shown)


aws-xray-recorder-sdk-core/src/main/java/com/amazonaws/xray/contexts/LambdaSegmentContext.java
```diff
@@ -50,25 +51,18 @@ private static boolean isInitializing(TraceHeader traceHeader) {
...
             logger.debug("Beginning subsegment named: " + name);
         }
 
+        TraceHeader traceHeader = LambdaSegmentContext.getTraceHeaderFromEnvironment();
         Entity entity = getTraceEntity();
-        if (entity == null) { // First subsgment of a subsegment branch.
-            Segment parentSegment = newFacadeSegment(recorder, name);
+        if (entity == null) { // First subsegment of a subsegment branch
+            Segment parentSegment = isInitializing(traceHeader)
+                ? Segment.noOp(TraceID.create(recorder), recorder)
```
wangzlei, line 64: can we add a debug log here, indicates the xray sdk creates no-op segment.

aws-xray-recorder-sdk-core/src/test/java/com/amazonaws/xray/contexts/LambdaSegmentContextTest.java
```diff
@@ -49,6 +50,8 @@ class LambdaSegmentContextTest {
 
     private static final String MALFORMED_TRACE_HEADER =
         ";;Root=1-57ff426a-80c11c39b0c928905eb0828d;;Parent=1234abcd1234abcd;;;Sampled=1;;;";
+    private static final String MALFORMED_TRACE_HEADER_2 =
```
wangzlei, line 53: Please add a test case "Root=1-5759e988-bd862e3fe1be46a994272711;Lineage=10:1234abcd:3", naming it as `Root Lambda passthrough`

wangzlei (approved):
LGTM

PR #404, opened 2024-05-28
[Lambda] Do not propagate trace header if dummy subsegment

majanjua-amzn:
Make it so that if the current entity is a dummy subsegment, we do not propagate the trace header from this. Instead, rely on AWS SDK to propagate the trace header in the environment. This makes it so that the trace ID of the context is appropriately propagated instead of an empty/default trace ID.

*Description of changes:*
- Update any additions of the trace header that rely on subsegments to check for no op subsegments
- Update tests accordingly


wangzlei (approved):
LGTM

PR #406, opened 2024-06-14
add xray sdk language and version in subsegment from Lambda environment

wangzlei:
*Issue #, if available:*
X-Ray Segments have X-Ray SDK and version information in field `aws: {xray {}}`(refer to [code](https://github.com/aws/aws-xray-sdk-java/blob/master/aws-xray-recorder-sdk-core/src/main/java/com/amazonaws/xray/AWSXRayRecorder.java#L502)), but in AWS Lambda environment there is only Subsegment can be generated. Customer does not know the xray sdk version by checking trace from xray console.

*Description of changes:*
Appending xray sdk info into the first subsegment generated in each Lambda invocation.

```
     aws": {
                    "xray": {
                        "sdk_version": "2.16.0",
                        "sdk": "X-Ray for Java"
                    }
                },
```

### DIFF
--- a/aws-xray-recorder-sdk-core/src/main/java/com/amazonaws/xray/contexts/LambdaSegmentContext.java
+++ b/aws-xray-recorder-sdk-core/src/main/java/com/amazonaws/xray/contexts/LambdaSegmentContext.java
@@ -80,6 +80,7 @@ public class LambdaSegmentContext implements SegmentContext {
                     ? new SubsegmentImpl(recorder, name, parentSegment)
                     : Subsegment.noOp(parentSegment, recorder);
             subsegment.setParent(parentSegment);
+            subsegment.putAllAws(recorder.getAwsRuntimeContext());
             // Enable FacadeSegment to keep track of its subsegments for subtree streaming
             parentSegment.addSubsegment(subsegment);
             setTraceEntity(subsegment);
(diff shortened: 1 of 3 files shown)


aws-xray-recorder-sdk-core/src/main/java/com/amazonaws/xray/contexts/LambdaSegmentContext.java
```diff
@@ -80,6 +80,7 @@ public Subsegment beginSubsegment(AWSXRayRecorder recorder, String name) {
                     ? new SubsegmentImpl(recorder, name, parentSegment)
                     : Subsegment.noOp(parentSegment, recorder);
             subsegment.setParent(parentSegment);
+            subsegment.putAllAws(recorder.getAwsRuntimeContext());
```
jj22ee, line 83: If user creates multiple "root subsegments", looks like each will have a copy of this information.
Do we want only one of the subsegments to have this information?
I don't know which is more reasonable: one or all subsegments with this data.

srprash: We are adding this information to all subsegments and not just "root subsegments" (not sure @jj22ee what you mean by this), right? 
So it seem redundant to add same information to all the subsegments. Can we possibly add this to the segment created by Lambda runtime?

PR #409, opened 2024-08-01
Support Lambda PassThrough trace header propagation

majanjua-amzn:
## Issue

Lambda is implementing PassThrough mode which allows customers to send traces through instrumented lambda functions without those functions being traced. The way this is done is by changing the trace header structure from `Root=...;Parent=...;Sampled=1` in the active tracing case to `Root=...` in the PassThrough case. As such, our SDK needs to be updated to handle this.

## Changes

- Update logic to use no-op segment if parent and sampled are missing (ie. PassThrough mode)
  - Had to use `SuppressWarnings("nullness")` because `TraceHeader.getRootTraceId()` returns an `@Nullable TraceId` but `Segment.noOp()` expects a `@NonNull TraceId`
- Update all trace header propagations to only propagate `Root` if we're not actively sampling
  - Not propagating Parent and Sampled is fine because Parent will be re-generated in the next traced context and missing Sampled will be treated as not actively sampling anyways
  - We can't propagate only one of these, it needs to be either both or neither
- Updated unit tests

## Testing

Ran many extensive tests using v2.16.0, v2.17.0, and these changes to compare the outcomes. In these tests I configured 3 lambda functions to interact with each other ALL instrumented with the new changes:
- Lambda A: Calls Lambda B
- Lambda B: Calls Lambda C
- Lambda C: Calls `GetAccountSettings`

***The trace headers in the below cases are from logging `System.getenv("_X_AMZN_TRACE_ID")` in the lambda functions themselves***

The following are some important cases that prove the functionality of the code change

### Case 1.a: Lambda A Active &rarr; Lambda B PassThrough &rarr; Lambda C Active
- Lineage is NOT propagated or received by any of these functions
- Lambda B has an inferred node due to the invocation in Lambda A which is actively traced, this is expected behaviour and not controlled by the SDK
- Lambda B function is being appropriately suppressed aside from the previous note
- Lambda C invocation is appropriately connected to the Lambda A function

```
A: Root=1-66abd797-167a78c02dcfeb4551890716;Parent=ad075ec25766d3f9;Sampled=1;Lineage=456f6a97:0
B: Root=1-66abd797-167a78c02dcfeb4551890716;Parent=57696313aea6f4a2;Sampled=1;Lineage=8408dcb0:0
C: Root=1-66abd797-167a78c02dcfeb4551890716;Parent=710d6393cf68bccc;Sampled=1;Lineage=7d2b6fd0:0
```
Image

### Case 1.b: Same thing but Lambda A Active sets `Sampled=0`
- Lineage is NOT propagated or received by any of the functions
- No trace is generated
- Lambda A, B, and C all have Sampled=0
- Lambda C appropriately changes Parent as expected, as it is actively traced, but still maintains the Sampled=0

```
A: Root=1-66b32340-6a4114ff44ca3e5240b1ac5b;Parent=63033870537134cd;Sampled=0;Lineage=456f6a97:0
B: Root=1-66b32340-6a4114ff44ca3e5240b1ac5b;Parent=63033870537134cd;Sampled=0;Lineage=8408dcb0:0
C: Root=1-66b32340-6a4114ff44ca3e5240b1ac5b;Parent=71d11ee442af5530;Sampled=0;Lineage=7d2b6fd0:0
```

### Case 2: Lambda A PassThrough &rarr; Lambda B Active &rarr; Lambda C PassThrough
- Lineage is NOT propagated or received by any of these functions
- Lambda B is the start of the trace as expected
- Lambda C invocation is still attached to Lambda B as an inferred node (similar to case 1)
- Lambda C's `GetAccountSettings` call IS TRACED; this is expected behaviour because we entered Active tracing mode in Lambda B and this is meant to propagate to downstream services. If the downstream service has its own version of passive tracing/X-Ray integration, it should be configured as such, but PassThrough functionality should only apply to Lambda functions and not the downstream calls they trigger unless the entire pathway has been in PassThrough
```
A: Root=1-66b3b030-7155e4f41eac6ff601dfaf63;Lineage=456f6a97:0
B: Root=1-66b3b030-7155e4f41eac6ff601dfaf63;Parent=ad139ed6effb18d2;Sampled=1;Lineage=8408dcb0:0
C: Root=1-66b3b030-7155e4f41eac6ff601dfaf63;Parent=50262dfb251d8754;Sampled=1;Lineage=7d2b6fd0:0
```
Image (1)

### Case 3.a: Calling Lambda B PassThrough &rarr; Lambda C PassThrough
- Lineage not propagated
- No trace, proves statement above that a complete trace entirely in PassThrough mode will not trace the final `GetAccountSettings` call since we never entered Active tracing mode.
```
B: Root=1-66abfc06-3cbd74727bb2f86a6a772af2;Lineage=8408dcb0:0
C: Root=1-66abfc06-3cbd74727bb2f86a6a772af2;Lineage=7d2b6fd0:0
```

### Case 3.b: Calling Lambda B Active &rarr; Lambda C Active
- Lineage not propagated
- Complete and correct trace
```
B: Root=1-66abfd06-5e95193a3dcfa6a25cec5b30;Parent=11c07dc8ed123be7;Sampled=1;Lineage=8408dcb0:0
C: Root=1-66abfd06-5e95193a3dcfa6a25cec5b30;Parent=613a4ff907527f30;Sampled=1;Lineage=7d2b6fd0:0
```
Screenshot 2024-08-01 at 2 25 21 PM

### DIFF
--- a/aws-xray-recorder-sdk-apache-http/src/main/java/com/amazonaws/xray/proxies/apache/http/TracedHttpClient.java
+++ b/aws-xray-recorder-sdk-apache-http/src/main/java/com/amazonaws/xray/proxies/apache/http/TracedHttpClient.java
@@ -18,7 +18,7 @@ package com.amazonaws.xray.proxies.apache.http;
 import com.amazonaws.xray.AWSXRay;
 import com.amazonaws.xray.AWSXRayRecorder;
 import com.amazonaws.xray.entities.Namespace;
-import com.amazonaws.xray.entities.NoOpSubSegment;
+import com.amazonaws.xray.entities.NoOpSegment;
 import com.amazonaws.xray.entities.Segment;
 import com.amazonaws.xray.entities.Subsegment;
 import com.amazonaws.xray.entities.TraceHeader;
@@ -107,7 +107,7 @@ public class TracedHttpClient extends CloseableHttpClient {
         if (subsegment.shouldPropagate()) {
             // If no-op, only propagate root trace ID to not taint sampling decision
             TraceHeader t = TraceHeader.fromEntity(subsegment);
-            if (subsegment instanceof NoOpSubSegment) {
+            if (parentSegment instanceof NoOpSegment) {
                 request.addHeader(
                         TraceHeader.HEADER_KEY,
                         "Root=" + t.getRootTraceId().toString());
--- a/aws-xray-recorder-sdk-apache-http/src/main/java/com/amazonaws/xray/proxies/apache/http/TracedHttpClient.java
+++ b/aws-xray-recorder-sdk-apache-http/src/main/java/com/amazonaws/xray/proxies/apache/http/TracedHttpClient.java
@@ -104,7 +104,16 @@ public class TracedHttpClient extends CloseableHttpClient {
         Segment parentSegment = subsegment.getParentSegment();
 
         if (subsegment.shouldPropagate()) {
-            request.addHeader(TraceHeader.HEADER_KEY, TraceHeader.fromEntity(subsegment).toString());
+            // If the trace is not sampled, passing Parent and Sampled won't matter
+            TraceHeader t = TraceHeader.fromEntity(subsegment);
+            if (t.getSampled() != TraceHeader.SampleDecision.SAMPLED) {
+                request.addHeader(
+                        TraceHeader.HEADER_KEY,
+                        "Root=" + t.getRootTraceId().toString());
+            } else {
+                // This will propagate Parent and Sampled
+                request.addHeader(TraceHeader.HEADER_KEY, TraceHeader.fromEntity(subsegment).toString());
+            }
         }
 
         Map<String, Object> requestInformation = new HashMap<>();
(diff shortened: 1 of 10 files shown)


aws-xray-recorder-sdk-apache-http/src/main/java/com/amazonaws/xray/proxies/apache/http/TracedHttpClient.java
```diff
@@ -104,7 +104,16 @@ public static void addRequestInformation(Subsegment subsegment, HttpRequest requ
         Segment parentSegment = subsegment.getParentSegment();
 
         if (subsegment.shouldPropagate()) {
-            request.addHeader(TraceHeader.HEADER_KEY, TraceHeader.fromEntity(subsegment).toString());
+            // If the trace is not sampled, passing Parent and Sampled won't matter
+            TraceHeader t = TraceHeader.fromEntity(subsegment);
+            if (t.getSampled() != TraceHeader.SampleDecision.SAMPLED) {
+                request.addHeader(
+                        TraceHeader.HEADER_KEY,
+                        "Root=" + t.getRootTraceId().toString());
```
alexey-katranov, line 112: I am a bit confused with this line, does it drop Parent and Sampled? I.e. if we have header `Root=...;Parent=...;Sampled=0` will we propagate only `Root=...`?
If so, it can cause undesired behaviour if the downstream in active tracing mode because `Root=...`-only header can cause sampling decision to be `1`. I.e. if the upstream decides not to sample then the downstream with active tracing can change the decision.

majanjua-amzn: Changed logic to detect passthrough mode via `NoOpSubSegment` instead of relying on sampling status.

This is because `SamplingDecision.UNKNOWN` is never the outcome of getting the trace header from the subsegment directly, so we need to rely on the type of the subsegment.

alexey-katranov (approved):
LGTM, however, I am not sure if Lineage is properly handled.

wangzlei (approved):

PR #428, opened 2025-07-30
Remove AWS SDK v1 Dependency from X-Ray SDK Core

lukeina2z:
The X-Ray SDK Core previously relied on AWS SDK v1 solely to support AWS Remote Sampling. However, the SDK does not use the SDK’s client to call the sampling service directly. Instead, it composes REST requests manually and sends them through an AWS proxy. The only part of AWS SDK v1 being used was the data model types for serialization into JSON.

This change removes the AWS SDK v1 dependency by reimplementing the necessary data types internally, eliminating the need to rely on the external SDK.

Test Plan:
Unit tests: Pass
E2E tests with X-Ray Daemon: Pass
E2E tests with CloudWatch Agent: Pass


aws-xray-recorder-sdk-core/src/main/java/com/amazonaws/xray/strategy/sampling/GetSamplingRulesResponse.java
srprash, line 1: Just curious: Is this and the other new data model classes based off of existing classes in AWS SDK or somewhere else?

aws-xray-recorder-sdk-core/src/main/java/com/amazonaws/xray/strategy/sampling/XRayClient.java
```diff
@@ -45,22 +30,7 @@ private XRayClient() {
...
-                DUMMY_REGION
-        );
-
-        return AWSXRayClientBuilder.standard()
-                .withEndpointConfiguration(endpointConfig)
-                .withClientConfiguration(clientConfig)
-                .withCredentials(ANONYMOUS_CREDENTIALS) // This will entirely skip signing too
-                .build();
-
+    public static Object newClient() {
```
srprash, line 33: Can even be fully removed?

build.gradle.kts
```diff
@@ -120,6 +120,12 @@ allprojects {
 
             // TODO(anuraaga): Enable on all projects.
             skipCheckerFramework = project.name != "aws-xray-recorder-sdk-core" || JavaVersion.current() != JavaVersion.VERSION_11
+
+            // Two reasons we cannot run the Nullness Checker on aws-xray-recorder-sdk-core:
+            // 1. Google AutoValue does not generate @Nullable or @NonNull annotations on methods like equals() and hashCode().
+            // 2. The existing codebase does not consistently perform null checks, in line with the behavior of the X-Ray SDK:
+            //    https://docs.aws.amazon.com/xray/latest/api/API_SamplingTargetDocument.html
+            skipCheckerFramework = skipCheckerFramework || project.name == "aws-xray-recorder-sdk-core"
```
srprash, line 128: This conflicts with the condition above where we skip null checker on all projects except the `aws-xray-recorder-sdk-core`. And now we are disabling on this project itself. What's even the point of having this checker now?

aws-xray-recorder-sdk-core/src/main/java/com/amazonaws/xray/strategy/sampling/pollers/RulePoller.java
```diff
@@ -96,8 +95,8 @@ private void pollRule() {
         Instant now = clock.instant();
 
         logger.info("Polling sampling rules.");
-        GetSamplingRulesRequest req = new GetSamplingRulesRequest();
-        GetSamplingRulesResult records = client.getSamplingRules(req);
+        GetSamplingRulesRequest req = GetSamplingRulesRequest.create(null);
```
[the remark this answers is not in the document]
lukeina2z: Here is the definition of create() method:  the parameter is @Nullable

    public static GetSamplingRulesRequest create(@Nullable String nextToken) {
        return new AutoValue_GetSamplingRulesRequest(nextToken);
    }

aws-xray-recorder-sdk-core/src/main/java/com/amazonaws/xray/strategy/sampling/XRayClient.java
lukeina2z: Yes, we can. Removing it actually makes my work simpler. However, by doing that, customer will potentially have to modify their code with the new release. So the reason I am keeping it here is customer/dev experience. We should not introduce a build break after they upgrade to the new version of the SDK.

build.gradle.kts
lukeina2z: Good point. I will remove it.

